PR #12, opened 2015-03-06
Better handling for deprecated functions

Sobak:
Description and complementary pull request for web-php is [here](https://github.com/php/web-php/pull/63)

### DIFF
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -116,13 +116,21 @@ class Package_PHP_Web extends Package_PHP_XHTML {
             if (!isset($written_toc[$filename])) {
                 $written_toc[$filename] = true;
 
-                $toc = array();
+                $toc = $toc_deprecated = array();
 
                 foreach($siblings as $sid) {
-                    $toc[] = array(
+                    $sibling_short_desc = Format::getShortDescription($sid);
+
+                    $entry = array(
                         Format::getFilename($sid).$ext,
-                        Format::getShortDescription($sid),
+                        $sibling_short_desc,
                     );
+
+                    if (isset($this->deprecated[$sibling_short_desc])) {
+                        $toc_deprecated[] = $entry;
+                    } else {
+                        $toc[] = $entry;
+                    }
                 }
 
                 $parents = array();
@@ -136,6 +144,7 @@ class Package_PHP_Web extends Package_PHP_XHTML {
 
                 $content = '<?php
 $TOC = ' . var_export($toc, true) . ';
+$TOC_DEPRECATED = ' . var_export($toc_deprecated, true) . ';
 $PARENTS = ' . var_export($parents, true) . ';';
 
                 file_put_contents($this->getOutputDir() . $filename, $content);
@@ -186,6 +195,7 @@ $PARENTS = array();
 '.$incl.'
 $setup = '.$var.';
 $setup["toc"] = $TOC;
+$setup["toc_deprecated"] = $TOC_DEPRECATED;
 $setup["parents"] = $PARENTS;
 manual_setup($setup);
 
--- a/phpdotnet/phd/Package/PHP/XHTML.php
+++ b/phpdotnet/phd/Package/PHP/XHTML.php
@@ -124,6 +124,7 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
 
     private $versions = array();
     private $acronyms = array();
+    protected $deprecated = array();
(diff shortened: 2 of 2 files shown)


phpdotnet/phd/Package/PHP/Web.php
```diff
@@ -116,13 +117,22 @@ public function header($id) {
...
+                $toc = $toc_deprecated = array();
 
                 foreach($siblings as $sid) {
-                    $toc[] = array(
-                        Format::getFilename($sid).$ext,
-                        Format::getShortDescription($sid),
-                    );
+                    $sibling_short_desc = Format::getShortDescription($sid);
+
+                    if (!isset($this->deprecated[$sibling_short_desc])) {
```
bjori: I prefer if (truth) over if (not true) expressions.

Please make it if(isset()) {.... } else {} :)

phpdotnet/phd/Package/PHP/Web.php
```diff
@@ -116,13 +117,22 @@ public function header($id) {
...
+                        $toc[] = array(
+                            Format::getFilename($sid).$ext,
+                            $sibling_short_desc,
+                        );
+                    } else {
+                        $toc_deprecated[] = array(
+                            Format::getFilename($sid).$ext,
+                            $sibling_short_desc,
+                        );
+                    }
```
bjori: To avoid future confusion, this should likely be
$entry = array(.....);
and the if (deprecated) { $toc_deprecated[] = $entry; } else { $toc[] = $entry; }

phpdotnet/phd/Package/PHP/Web.php
```diff
@@ -73,6 +73,7 @@ public function update($event, $val = null) {
 
         case Render::INIT:
             $this->loadVersionAcronymInfo();
+            $this->deprecated = $this->generateDeprecatedInfo(Config::phpweb_version_filename());
```
bjori: Shouldn't this really be added to loadVersionAcronymInfo() ?

phpdotnet/phd/Package/PHP/Web.php
```diff
@@ -211,6 +223,40 @@ protected function writeJsonIndex() {
         v("Index written", VERBOSE_FORMAT_RENDERING);
     }
 
+    protected static function generateDeprecatedInfo($filename) {
+        static $info;
```
bjori: It seems like this function can (should?) be merged with generateVersionInfo() and both be done at the same time, rather the iterating over the file twice

Sobak: I'm also for avoiding code duplication and being DRY, but it would mean that we fetch these `$deprecated` data for XHTML format as well. Are we okay with that? I'll fix rest of the issues soon.

Sobak: After reconsidering what you said, I have mixed feelings. `generateVersionInfo()` should return array unless we don't want to insert deprecated deprecation state directly into `$this->versions`, making it multidimensional array.

It would also require second condition, either way, because when generating indices for `$this->depracted` I need to replace special characters with underscores, not hyphens. Otherwise, I wouldn't be able to compare them with result of `Format::getShortDescription()`.

```php
// generateVersionInfo()
$funcname = str_replace(
array("::", "->", "__", "_", '$'),
array("-",  "-",  "-",  "-", ""),
$r->value)

// generateDeprecatedInfo()
$funcname = str_replace(
array("::", "->", "__", "_", '$'),
array("_",  "_",  "_",  "_", ""),
$r->value)
```

What do you think?

phpdotnet/phd/Package/PHP/Web.php
morrisonlevi: Agreed; the work is the same and the duplication should be avoided.

phpdotnet/phd/Package/PHP/Web.php
bjori: Ahhh. I see how versionInfo() does it now. That seems like we could have simplified this and safed tons of method calls.
I suppose at this time its not worth the troubles to fix it.


I still prefer to move this method to the same place, and I'm fine with loading the file rendering the pure html output too even though its not used.

bjori: (and maybe make an inline note of the different str_replace() - thats not easy to spot and someone could accidentally merge these functions without noticing)

Sobak:
@bjori @morrisonlevi I think I fixed all the issues you guys pointed out. Thank you for your time!

bjori:
lgtm

PR #15, opened 2017-08-17
Keep the return value to something count-able

tpunt:
In PHP 7.2, `count` can only be used with arrays and `Countable` objects. Returning `null` here causes a lot of warnings when building the docs on PHP 7.2.


salathe:
Thanks for the update, Thomas. 👍

PR #18, opened 2018-01-31
Update README

royopa:


philip:
Rodrigo, thanks for the patch :)

PR #20, opened 2019-02-10
Correct continue usage within a switch block

duncan3dc:
`continue` within a `switch` block works exactly the same as `break` and is normally a mistake. There's no code after the switch so presumably `continue 2` was not intended here, so this PR just changes it to `break` to avoid warnings in PHP 7.3


petk:
Applied. Thank you @duncan3dc

PR #21, opened 2019-12-17
changed join parameter order for PHP 7.4.0

mumumu:
When I use PhD with PHP 7.4.0, I got the following warning flood.

```
        join(): Passing glue string after array is deprecated. Swap the parameters
[01:24:28 - E_DEPRECATED          ] /home/mumumu/.phpenv/versions/7.4.0/lib/php/phpdotnet/phd/Package/Generic/Manpage.php:630
```

Fixed the changed join parameter order, which warning suggests.


salathe:
Thanks @mumumu 👍

PR #22, opened 2020-03-04
Add support for properly rendering nullable types

cmb69:
While it is already possible to write types with the nullable type mark
(`?`) (and some extensions already make use of this), these types are
not linked to the respective classes or interfaces.  So we make sure
that the type lookup ignores the question mark, and we put it outside
the link.

This is a bit hackish, because it would be better to introduce proper
support for union types; however, these are only supported as of
DocBook 5.2, and since we're still using 4.5, we would have to upgrade
first.

---

Looks like this:

phd-nullable

### DIFF
--- a/phpdotnet/phd/Package/PHP/XHTML.php
+++ b/phpdotnet/phd/Package/PHP/XHTML.php
@@ -420,13 +420,20 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
     }
 
     public function format_type_if_object_or_pseudo_text($type, $tagname) {
-        if (in_array(strtolower($type), array("bool", "int", "double", "boolean", "integer", "float", "string", "array", "object", "resource", "null"))) {
+        $mandatory = $type[0] === '?' ? substr($type, 1) : $type;
+        if (in_array(strtolower($mandatory), array("bool", "int", "double", "boolean", "integer", "float", "string", "array", "object", "resource", "null"))) {
             return false;
         }
         return self::format_type_text($type, $tagname);
     }
 
     public function format_type_text($type, $tagname) {
+        if ($type[0] === '?') {
+            $prefix = '?';
+            $type = substr($type, 1);
+        } else {
+            $prefix = '';
+        }
         $t = strtr(strtolower($type), "_", "-");
         $href = $fragment = "";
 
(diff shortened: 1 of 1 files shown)


cmb69:
Any objections, or suggestions for improvement? /cc @rquadling

rquadling:
Looking at the options for `<type>` in DocBook, the only available attribute is `role` (
https://tdg.docbook.org/tdg/5.1/type.html & https://tdg.docbook.org/tdg/5.1/ref-elements.html#common.attributes)

The value of the role is soft and so we would need to define roles. Of course, the current one would be 'nullable', but could easily be a CSL of roles.

If we consider nullable types as being a different type, then this makes sense. If not, and the nullable aspect is about the nature of the value of the argument, rather than its type, then the type should not be amended (i.e. prefixed with `?`).

As we already have https://tdg.docbook.org/tdg/5.1/methodparam.html / `<methodparam choice="opt">` to indicate that a parameter is optional. I think we should be putting the nullable aspect to the parameter here in some suitable form. And role on methodparam is (having written all of this) probably where I'd define the nullable-ness of the parameter. Not the type as the type doesn't actually change.

Role seems to be pretty much a free for all usage wise. 

Also, we need to make sure that the parameters in question are actually nullable (i.e. you can supply null and the method/function works as expected) as this is one of the simplest mistakes I've seen in my team thinking that they can swap to nullable types and not use a null default. They are different things. 

One is a mandatory param that MAY be null: `<methodparam role="nullable"><type>string</type><parameter>mandatory_param_that_can_be_a_string_or_null</parameter></methodparam>`.

The other is an optional param and will default to null: `<methodparam choice="opt"><type>string><parameter>optional_string_that_defaults_to_null</parameter><initializer>&null;</initializer></methodparam>`.

Of course, it may be nullable AND optional: `<methodparam role="nullable"><type>string</type><parameter>mandatory_param_that_can_be_a_string_or_null</parameter><initializer>&null;</initializer></methodparam>`

cmb69:
Using the `role` attribute is intriguing, thanks! However, I'm not convinced that this would be really suitable, since it is not only about parameters, but also return types and types of properties, for example. Furthermore, a nullable type (e.g. `?string`) is actually a shortcut for a union type (e.g. `string|null`), and this could be expressed with DocBook 5.2 as `<type class="union"><type>string</type><type>null</type></type>`. Until we could use that DocBook version, it seems to me that actually using a `?` prefix is most suitable.

Regarding the difference of nullable and optional parameters: I'm aware of that. :) This PR is to support php/doc-en#54 which is indeed about some mandatory nullable parameters (which made no sense when they had been introduced, because the behavior of passing an empty string and NULL are identical, but we cannot remove the nullability anymore (without breaking BC) due to strict typing.

cmb69:
Ah, well, yet another option would be to *render* nullable types as union, or to even write them in the docs as such.

cmb69:
> Ah, well, yet another option would be to render nullable types as union, or to even write them in the docs as such.

See PR #25.

cmb69:
I'm closing this PR in favor of PR #25.

PR #23, opened 2020-03-05
Support namespaced classes in <type> elements

cmb69:
Currently, automatic linking to the documentation pages of namespaced
classes is supported in some elements (e.g. in `<classsynopsisinfo>`),
but not in `<type>` elements, which can be seen in the `CommonMark`
documentation[1].  We catch up on that.

[1] <https://www.php.net/manual/en/book.cmark.php>

---

@krakjoe, you may like it:
commonmark-node-methods


cmb69:
If there are no objections, I'll merge this in a week.

cmb69:
Applied as 3177809ebd0cac16708321f5f226b11876e1a948. Thanks.

PR #24, opened 2020-03-07
Improve parameter rendering

simPod:
_This is a reaction to https://github.com/php/doc-en/pull/55_

Allows to properly render:
- arguments
- arguments by reference
- varible-length argument lists
- varible-length argument lists passed by reference

----------

It will also require changes (see below) in docs repository (I will send PR) but did not find a non-BC breaking way because:

- params in docs are currently named as `...`
- the param name build is performed as `<tag>$` + `param_name` resulting in `<tag>$param_name`
- however, that does not work for variadic args as we need `...$args` instead of `$...args`

-----------

Current:

```xml
<methodparam choice="opt">
    <type>bool|float|int|string</type>
    <parameter>...</parameter>
</methodparam>
```

`sprintf ( string $format [, mixed $... ] ) : string`

This PR allows:

```xml
<methodparam choice="opt">
    <type>bool|float|int|string</type>
    <parameter role="vararglist">args</parameter>
</methodparam>
```

`sprintf ( string $format [, mixed ...$args ] ) : string`

### DIFF
--- a/phpdotnet/phd/Package/Generic/Manpage.php
+++ b/phpdotnet/phd/Package/Generic/Manpage.php
@@ -583,10 +583,29 @@ class Package_Generic_Manpage extends Format_Abstract_Manpage {
     }
 
     public function format_parameter_method($open, $name, $attrs, $props) {
-        if ($open && isset($attrs[Reader::XMLNS_DOCBOOK]["role"]) && $attrs[Reader::XMLNS_DOCBOOK]["role"] == "reference") {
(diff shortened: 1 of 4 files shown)


cmb69:
Oh, I haven't seen this PR before. From what I can tell, this is basically the same as PR #28, while potentially #29 would be preferable.

simPod:
Kk let's proceed there

cmb69:
PR #29 has just been merged; it seems to me this PR is now obsolete. However, we may need to port the changes to other renderers, although I'm not sure whether these are still used at all.

PR #25, opened 2020-03-15
Support union types

cmb69:
For now we stick with the union type syntax introduced with PHP 8,
because proper union type markup[1] is only available as of DocBook
5.2, and we're still using version 4.5.  We do not support the nullable
type syntax (i.e. ? prefix) which has been introduced with PHP 7.1,
because this is just a shortcut for union types, and may not be
supported by future DocBook versions.

[1] <https://tdg.docbook.org/tdg/5.2/type.html>

---

This is an alternative to PR #22, which may not have been the best idea, although that PR would support some parts of the documentation which already use nullable types; these would have to be changed to union types.

Anyhow, union types would be rendered like this:

union-types

### DIFF
--- a/phpdotnet/phd/Package/PHP/XHTML.php
+++ b/phpdotnet/phd/Package/PHP/XHTML.php
@@ -420,10 +420,15 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
     }
 
     public function format_type_if_object_or_pseudo_text($type, $tagname) {
-        if (in_array(strtolower($type), array("bool", "int", "double", "boolean", "integer", "float", "string", "array", "object", "resource", "null"))) {
-            return false;
+        $res = [];
+        foreach (explode('|', $type) as $t) {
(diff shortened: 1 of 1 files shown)


cmb69:
If there are no objections, I'll merge this in a week.

salathe:
👎  - we should upgrade DocBook and use the "new" union types, or backport them into our current not-quite-standard DocBook DTD, and change PhD to nicely render those types.

cmb69:
Objection sustained! :) (I'm leaving this PR open for now, as a reminder)

Any volunteers for doing it properly?

deleugpn:
I'm going to start investigating what I need to learn to get this up. Hints are welcomed!

cmb69:
Upgrading our customized docbook might require a major effort; backporting union type support (see https://github.com/docbook/docbook/issues/91) may make more sense for now. That would have to be done for https://github.com/php/doc-en, and would be a prerequisite for actually supporting union types in PhD.

cmb69:
Closing in favor of PR #30.

PR #29, opened 2020-09-13
Support variadic parameters – take 2

cmb69:
So far, PhD does not support variadic parameters.  Instead the manual
follows the convention to use `...` as parameter name, which is then
rendered as `$...` in the docs.  However, as of PHP 5.6.0 variadic
parameters are supported using the syntax `...$args`; therefore, we are
going to support this notation for PhD.

While DocBook supports a `<varargs>` element, this cannot be used in
combination with named parameters.  Therefore, we use the `rep`
attribute of the `<methodparam>` element to designate variadic
parameters.

---

What I don't like about this solution is the semantic mismatch. Consider, e.g. `array_merge(array ...$arrays)`; it is not actually `$arrays` which is repeatable, but rather `$array`. Maybe I'm too nit-picky here.

TODO:
- [ ] adapt non XHTML based formats
- [ ] fix parameter linking for web-php

### DIFF
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -976,7 +976,7 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
 
     public function format_methodsynopsis($open, $name, $attrs) {
         if ($open) {
-            $this->params = array("count" => 0, "opt" => 0, "content" => "", "repeatable" => '');
+            $this->params = array("count" => 0, "opt" => 0, "content" => "", "ellipsis" => '');
             $id = (isset($attrs[Reader::XMLNS_XML]["id"]) ? ' id="'.$attrs[Reader::XMLNS_XML]["id"].'"' : '');
             return '<div class="'.$name.' dc-description"'.$id.'>';
         }
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -976,7 +976,7 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
 
     public function format_methodsynopsis($open, $name, $attrs) {
         if ($open) {
-            $this->params = array("count" => 0, "opt" => 0, "content" => "");
+            $this->params = array("count" => 0, "opt" => 0, "content" => "", "repeatable" => '');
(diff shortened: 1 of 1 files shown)


cmb69:
Ugh, the PDF renderer seems to rely on [PECL/haru](https://pecl.php.net/package/haru), which is abandoned; latest release almost 8 years ago, and unlikely to be compatible with PHP 7.

cmb69:
Also, if we go this way, should `rep="repeat"` replace `choice="opt"`, or should it be used in addition?

phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -976,7 +976,7 @@ public function format_fieldsynopsis_modifier_text($value, $tag) {
 
     public function format_methodsynopsis($open, $name, $attrs) {
         if ($open) {
-            $this->params = array("count" => 0, "opt" => 0, "content" => "");
+            $this->params = array("count" => 0, "opt" => 0, "content" => "", "repeatable" => '');
```
nikic, line 979: Mismatch between use of `"repeatable"` here an `"ellipsis"` below.

cmb69: Thanks!

nikic:
> fix parameter linking for web-php

What's the problem there?

cmb69:
That should be solved with https://github.com/php/web-php/pull/349.

phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -1051,6 +1051,11 @@ public function format_methodparam($open, $name, $attrs) {
                 }
                 $content .= ' <span class="methodparam">';
                 ++$this->params["count"];
+                if (isset($attrs[Reader::XMLNS_DOCBOOK]["rep"]) && $attrs[Reader::XMLNS_DOCBOOK]["rep"] == "repeat") {
+                    $this->params["ellipsis"] = '...';
+                } else {
+                    $this->params["ellipsis"] = '';
+                }
```
nikic, line 1058: Why do we set `params["ellipsis"]` here rather than handling it directly in `format_methodparam_parameter`, like the `role` is?

cmb69: `role="reference"` is an attribute of `<parameter>`, but `rep="repeat"` is an attribute of `<methodparam>`.

nikic: Oh, duh.

nikic (approved):

cmb69:
Applied as d59932c1518cdc0daa92aac5ca1905b83475c6e6.

PR #30, opened 2020-10-01
Properly support union types

cmb69:
This is the cleaner alternative to PR #25, which supports nested
`<type>` elements, and renders them as union types in PHP 8 syntax.

---

This is just a quick and dirty POC for now, in the hope to get that going. Assuming our docbook would support nested types, the following patch to doc-en
````.diff
Index: strpos.xml
--- strpos.xml	(revision 350727)
+++ strpos.xml	(working copy)
@@ -11,7 +11,7 @@
   <methodsynopsis>
    <type>int</type><methodname>strpos</methodname>
    <methodparam><type>string</type><parameter>haystack</parameter></methodparam>
-   <methodparam><type>mixed</type><parameter>needle</parameter></methodparam>
+   <methodparam><type class="union"><type>string</type><type>int</type></type><parameter>needle</parameter></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>offset</parameter><initializer>0</initializer></methodparam>
   </methodsynopsis>
   <para>
````
would be rendered like
Screenshot 2020-10-01 133429
Note that the predefined types are now rendered as links, which might not be the worst idea generally.

### DIFF
--- a/phpdotnet/phd/Package/PHP/XHTML.php
+++ b/phpdotnet/phd/Package/PHP/XHTML.php
@@ -154,8 +154,8 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
         "refsynopsisdiv"               => null,
     );
 
-    /* Support for union types */
-    private $types = null;
+    /** @var int|null Number of already formatted types in the current union type */
+    private $num_types = null;
 
     protected $pihandlers = array(
         'dbhtml'        => 'PI_DBHTMLHandler',
@@ -313,15 +313,15 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
         $retval = '';
         if ($open) {
             if (isset($attrs[Reader::XMLNS_DOCBOOK]["class"])) {
-                $this->types = 0;
-            } elseif (isset($this->types)) {
-                if (!$is_return_type && $this->types > 0) $retval .= '|';
-                $this->types++;
+                $this->num_types = 0;
+            } elseif (isset($this->num_types)) {
+                if (!$is_return_type && $this->num_types > 0) $retval .= '|';
+                $this->num_types++;
             }
             if (!$is_return_type) $retval .= '<span class="type">';
         } else {
             if (isset($attrs[Reader::XMLNS_DOCBOOK]["class"])) {
-                $this->types = null;
+                $this->num_types = null;
             }
             if (!$is_return_type) $retval .= '</span>';
         }
--- a/phpdotnet/phd/Package/PHP/XHTML.php
+++ b/phpdotnet/phd/Package/PHP/XHTML.php
@@ -70,10 +70,10 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
         ),
         'type'                  => array(
             /* DEFAULT */          'format_type',
-            'methodsynopsis'    => 'format_methodsynopsis_type',
+            'methodsynopsis'    => 'format_suppressed_tags',
             'type'              => array(
                 /* DEFAULT */       'format_type',
-                'methodsynopsis' => 'format_methodsynopsis_type',
+                'methodsynopsis' => 'format_suppressed_tags',
             ),
         ),
         'varname'               => array(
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by salathe** [comment]
Thanks for pursuing this @cmb69.

> Note that the predefined types are now rendered as links, which might not be the worst idea generally.

It would be good to see a given type rendered the same regardless of whether it is a standalone type or part of a union type in the prototype.  

At least for the short term, make sure to pass the type through `Package_PHP_XHTML::format_type_if_object_or_pseudo_text()` for consistency.   That should just be a matter of adding `'type' => 'format_type_if_object_or_pseudo_text',` to `Package_PHP_XHTML::$mytextmap['type']`. 

I don't think I'd object if that method was phased out (and so all types became links), but that's probably a task best left for a later date (or at least a separate PR).

> Assuming our docbook would support nested types...

I've patched our old DocBook DTD with a quick change to support union types (not intersection) in php/doc-base@82d74e69032e89cd6c8f624f7baf9780c639ec71 so the above `strpos()` example should happily `configure.php` now.

File: phpdotnet/phd/Package/PHP/XHTML.php [hunk 1]
```diff
@@ -128,6 +128,8 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
     /* Current Chunk settings */
     protected $cchunk          = array();
     /* Default Chunk settings */
```
salathe, line 130: This comment is now in the wrong place. It belongs on the `$dchunk` property.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -128,6 +128,8 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
...
+    private $types = null;
+    /* Support for compound types */
```
salathe, line 132: Comment before code, please. :)

phpdotnet/phd/Package/PHP/XHTML.php
cmb69: Oops.

phpdotnet/phd/Package/PHP/XHTML.php
cmb69: Done. :)

cmb69:
> I don't think I'd object if that method was phased out (and so all types became links), but that's probably a task best left for a later date (or at least a separate PR).

I agree that this should kept separate. It might be useful to do that in advance, though, since it would simplify this PR; currently, the link to built-in types are only suppressed in methodsynopses, but not elsewhere.

Anyhow, checking https://github.com/php/doc-en/pull/151 shows that there's more work to be done (at least union return types are not properly rendered yet).

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -69,8 +69,12 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
...
         ),
         'type'                  => array(
-            /* DEFAULT */          'span',
-            'methodsynopsis'    => 'format_suppressed_tags',
+            /* DEFAULT */          'format_type',
+            'methodsynopsis'    => 'format_methodsynopsis_type',
+            'type'              => array(
+                /* DEFAULT */       'format_type',
+                'methodsynopsis' => 'format_methodsynopsis_type',
+            ),
```
cmb69, line 77: I just noticed that the distinction between `format_type()` and `format_methodsynopsis_type()` is useless (both are doing the same). I'll fix this later today.

cmb69: Nope, nonsense.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -297,6 +309,30 @@ public function autogenVersionInfo($refnames) {
...
+                $this->types++;
+            }
+            if (!$is_return_type) $retval .= '<span class="type">';
+        } else {
+            if (isset($attrs[Reader::XMLNS_DOCBOOK]["class"])) {
+                $this->types = null;
+            }
+            if (!$is_return_type) $retval .= '</span>';
+        }
+        return $retval;
```
nikic, line 328: I find the treatment of return types confusing here. Doesn't this effectively reduce to `if ($is_return_type) return '';` at the top of the function? In which case there's no point for a shared do_format_type method anymore.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -145,6 +154,9 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
         "refsynopsisdiv"               => null,
     );
 
+    /* Support for union types */
+    private $types = null;
```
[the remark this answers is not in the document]
cmb69: Yeah, that's PhD. :) But a good idea to rename the propery and add a more meaningful docblock, at least.

phpdotnet/phd/Package/PHP/XHTML.php
cmb69: `do_format_type()` does two things: a) it updates `$num_types` and b) it calculates the `$retval` for `$is_return_type`. Splitting the function into two functions would require to duplicate the code logic; not sure if that would be a simplification.

nikic: Why do we need num_types for the return types case though?

cmb69: Actually, we don't. Good catch, thanks!

nikic (approved):

cmb69:
Applied as https://github.com/php/phd/commit/4d91cb2e051243a6be095d12d16ad471111d61f6.

PR #31, opened 2020-10-07
Fix broken rendering of varnames in fieldsynopsis

cmb69:
A recent commit suppressed the tag formatting of `<varname>` elements
for `Package_PHP_XHTML`; this must not be done for `<varname>`s inside
of `<fieldsysnopsis>`, though.

---

The https://www.php.net/manual/en/mysqli.error.php for an example of the currently broken rendering.


salathe (approved):
LGTM 👍 

before and after

cmb69:
Applied as 7624cceec01bf5b659da53a53c826b73c569e053.

PR #32, opened 2020-11-02
Don't render "void" in empty parameter lists

cmb69:


nikic (approved):

cmb69:
Applied as 5405e79b75a8d98da0c279d9c8d2741ced60923f.

PR #33, opened 2020-11-13
Generate correct links for <type> elements

Girgias:


cmb69:
Could you please rebase?

Girgias:
Done the rebase and did some clean-up for the support at the same time (and add `static` although unlikely).

I also don't really understand the purpose of the `format_type_if_object_or_pseudo_text()` function.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -487,6 +487,10 @@ public function format_type_text($type, $tagname) {
         case "double":
             $href = "language.types.float";
             break;
+        // old name for callable
+        case "callback":
+            $href = "language.types.callable";
+            break;
```
cmb69, line 493: This had deliberately been removed. It seems the reintroduction is accidential.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -500,6 +504,8 @@ public function format_type_text($type, $tagname) {
             $href = "language.types.$t";
             break;
         case "mixed":
+        case "void":
+        case "static":
```
cmb69, line 508: ditto

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -500,6 +504,8 @@ public function format_type_text($type, $tagname) {
             $href = "language.types.$t";
             break;
         case "mixed":
+        case "void":
```
kocsismate, line 507: There's a `format_void()` methodas well, which apparently renders void return types.

kocsismate: Also, I added 2 `<span class="type">?</span>` tags yesterday, which should be updated as well

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -442,7 +442,7 @@ public function format_methodsynopsis($open, $name, $attrs, $props) {
         if ($this->cchunk["methodsynopsis"]["returntypes"]) {
             $types = [];
             foreach ($this->cchunk["methodsynopsis"]["returntypes"] as $return_type) {
-                $formatted_type = self::format_type_if_object_or_pseudo_text($return_type, "type");
```
kocsismate, line 445: I believe `phpdotnet\phd\Package_PHP_XHTML::format_type_if_object_or_pseudo_text` is not used anywhere else so itcan be removed

phpdotnet/phd/Package/PHP/XHTML.php
Girgias: Hum, I don't really know how to do this frankly, because from my understanding this works on the XML? So I would imagine it gets initially rendered with links like `A|null` before getting transformed again?

And I'll drop void then, didn't know it was already handled

phpdotnet/phd/Package/PHP/XHTML.php
Girgias: Indeed will drop this

phpdotnet/phd/Package/PHP/XHTML.php
kocsismate: yeah, phd converts XML to HTML, but in case of simple nullable types, I had to discard the render of the <type>null</type> subtype, and insert a `<span class="type">?</span>` before the other subtype of the union type gets rendered. So what I wanted to ask you is to add the href directly to these HTML tags as well.

Girgias: Can you point me to where to the name of this function?

kocsismate: `format_type()` and `format_methodsynopsis()` in PHP/XHTML.php

Girgias: `format_type()` doesn't call `format_type_text()` so I didn't update it as I imagine this is already not displaying links.

kocsismate: `format_type()` is only called for rendering the tag itself, but `format_type_text()` is automatically called for the text inside the type tags. Now, that you changed `format_methodsynopsis()`, `?` has the link at the return type positions, but since `format_type_text()` was left out, `?` at parameter positions`are missing the links.

phpdotnet/phd/Package/PHP/XHTML.php
kocsismate: There's still another problem with simple nullable parameters:
```
/* Methods */
public __construct(
    ? $locale,
    int $dateType,
    int $timeType,
    IntlTimeZone|DateTimeZone|string|null $timezone = null,
    IntlCalendar|int|null $calendar = null,
    string $pattern = ""
)
```

phpdotnet/phd/Package/PHP/XHTML.php
kocsismate: There's still another problem with simple nullable parameters:
```
/* Methods */
public __construct(
    ? $locale,
    int $dateType,
    int $timeType,
    IntlTimeZone|DateTimeZone|string|null $timezone = null,
    IntlCalendar|int|null $calendar = null,
    string $pattern = ""
)
```

Girgias: Frankly I do not understand how the fuck these method call happen, so I have no clue if what I'm doing makes any sense. When I opened this PR initially it was just after a commit which broke the generation of the links, so I looked at this and tried to fix it. But I have ZERO understanding of PhD, and just trying to figure out how shit works here is... painful.

kocsismate: It's a bit better now: `?` characters both at the parameter as well as the return type position have the link. However, neither `void`, not `false` have them. As mentioned previously, you can add the link directly to the former one at phpdotnet/phd/Package/Generic/XHTML.php:1045, but I'm not sure what to do with `false`.

In order to fix the issue with the disappearing type in case of simple nullable types, you have to apply the following patch:
```
Index: phpdotnet/phd/Package/PHP/XHTML.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
diff --git a/phpdotnet/phd/Package/PHP/XHTML.php b/phpdotnet/phd/Package/PHP/XHTML.php
--- a/phpdotnet/phd/Package/PHP/XHTML.php	(revision e64f1141d524e7ad56e0a88c5cf161cfe5a6341d)
+++ b/phpdotnet/phd/Package/PHP/XHTML.php	(date 1627560074288)
@@ -546,12 +546,14 @@
         case "array":
         case "object":
         case "resource":
+        case "callable":
+        case "iterable":
+            $href = "language.types.$t";
+            break;
         case "null":
             if ($this->simple_nullable) {
                 return "";
             }
-        case "callable":
-        case "iterable":
             $href = "language.types.$t";
             break;
         case "mixed":
```

PR #34, opened 2020-11-19
Fix #79747: __autoload not labeled as "Deprecated"

cmb69:
Although that info is available in version.xml, PhD does not recognize
it, because several special characters are replaced when putting the
function name into the array, but we missed to replace them back when
comparing with the actual function name.

### DIFF
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -125,7 +125,7 @@ class Package_PHP_Web extends Package_PHP_XHTML {
                         $sibling_short_desc,
                     );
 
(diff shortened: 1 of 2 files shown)


cmb69:
Any objections to merge this?

nikic (approved):
Might be worthwhile to factor out some of the code, i.e. use the same function when generating and fetching to avoid discrepancies. The deprecation and version code is also very similar, just with `-` vs `_`...

cmb69:
Good idea, but with a quick shot, I didn't manage to get any version/deprecation info. There is also similar code for other renderers, so this needs more time. For now I'm going with this quickfix, to resolve the actual issue.

PR #35, opened 2020-12-01
fix: Update phd-guide Because the old repo is not available.

bolechen:
change SVN to GIT


cmb69 (commented):
Thank you for the PR! The repo link should be changed, but otherwise this looks good to me.

salathe:
Applied via 01505c7

PR #37, opened 2021-01-03
Don't render optional parameters in brackets

cmb69:
In PHP, parameters with a default value are optional, so there is no
need to enclose these in brackets.  For methodparams which do not have
an initializer, we use a question mark instead.

---

Before:
before

After:
after

This appears to be the simplest implementation; if we wanted, we could stick with the brackets for the generic renderer, and only drop these for the PHP renderer. And of course, the choice of the question mark for missing initializers is arbitrary and could be change; however, it would be preferable to have initializers for all functions anyway.

/cc @kocsismate

PS: if this PR gets accepted, we also have to adapt https://www.php.net/manual/en/about.prototypes.php accordingly.

### DIFF
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -976,14 +976,11 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
 
     public function format_methodsynopsis($open, $name, $attrs) {
         if ($open) {
-            $this->params = array("count" => 0, "opt" => 0, "content" => "", "ellipsis" => '');
+            $this->params = array("count" => 0, "opt" => false, "init" => false, "content" => "", "ellipsis" => '');
             $id = (isset($attrs[Reader::XMLNS_XML]["id"]) ? ' id="'.$attrs[Reader::XMLNS_XML]["id"].'"' : '');
             return '<div class="'.$name.' dc-description"'.$id.'>';
         }
         $content = "";
-        if ($this->params["opt"]) {
-            $content = str_repeat("]", $this->params["opt"]);
-        }
         $content .= " )";
 
         $content .= "</div>\n";
@@ -1007,6 +1004,7 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
 
     public function format_initializer($open, $name, $attrs) {
         if ($open) {
+            $this->params["init"] = true;
             return '<span class="'.$name.'"> = ';
         }
         return '</span>';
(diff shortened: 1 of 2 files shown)


kocsismate:
Very nice! I was trying to implement something like that a few weeks ago, but I didn't get through it. I think the `?` default value is okeyish, at least I don't have a better idea :) Maybe `UNKNOWN` could also work? It's too long though...

P.S.: what about also removing the whitespace before `,` in this PR if there is no brackets anymore? :) In the meanwhile, I'll continue with the methodsynopsis generation so that we'll have less unknown defaults.

cmb69:
`UNKNOWN` was my favorite for quite some time, but it may be mistaken for an actually existing constant, so I came up with the question mark.

I'd leave the whitespace changes to a separate PR; I think these make sense with or without this one.

phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -1040,11 +1038,9 @@ public function format_methodparam($open, $name, $attrs) {
                     $content .= " (";
                 }
                 if (isset($attrs[Reader::XMLNS_DOCBOOK]["choice"]) && $attrs[Reader::XMLNS_DOCBOOK]["choice"] == "opt") {
-                    $this->params["opt"]++;
-                    $content .= "[";
+                    $this->params["opt"] = true;
                 } else if($this->params["opt"]) {
-                    $content .= str_repeat("]", $this->params["opt"]);
-                    $this->params["opt"] = 0;
+                    $this->params["opt"] = false;
```
nikic, line 1043: Just an `else` would be clearer imho. We can unconditionally overwrite it, rather than only if it is not already false.

nikic (approved):
This looks sensible to me. I think the `?` placeholder is okay as long as the goal is not to have it -- either by having a proper default value, or by splitting the signature into two overloaded signatures.

phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -1058,6 +1054,10 @@ public function format_methodparam($open, $name, $attrs) {
                 }
                 return $content;
         }
+        if ($this->params["opt"] && !$this->params["init"]) {
+            return " = ?</span>";
+        }
```
junichi11, line 1059: @cmb69 Is it possible to change `return " = ?</span>"` to `return "<span class="initializer"> = ?</span></span>"`?

cmb69: Ah, good point. Has `.initializer` already style rules?

junichi11: Other default values are surrounded with it. e.g.
```
<span class="methodname"><strong>apcu_dec</strong></span>
    ( <span class="methodparam"><span class="type">string</span> <code class="parameter">$key</code></span>
   , <span class="methodparam"><span class="type">int</span> <code class="parameter">$step</code><span class="initializer"> = 1</span></span>
   , <span class="methodparam"><span class="type">bool</span> <code class="parameter reference">&$success</code> = ?</span>
   , <span class="methodparam"><span class="type">int</span> <code class="parameter">$ttl</code><span class="initializer"> = 0</span></span>
   ) : <span class="type"><span class="type">int</span>|<span class="type"><span class="type false">false</span></span></span></div>
```
https://www.php.net/manual/en/function.apcu-dec.php

cmb69: Ah, right, of course! Thanks for pointing this out. Done with c1dbe7b009e9399fa1024df963266ee52d365d62.

junichi11: No problem :) Thanks a lot for your quick fix!

PR #39, opened 2021-01-27
Fixed E_WARNING "($context) must be passed by reference, value given".

mumumu:
When I use PhD with PHP 8.0.x, I get the following warning flood.

```
[17:55:07 - E_WARNING             ] /home/mumumu/.phpenv/versions/8.0.1/lib/php/phpdotnet/phd/Format.php:150
        phpdotnet\phd\Format::SQLiteExample(): Argument #1 ($context) must be passed by reference, value given
```

I can't find why this warning occurs, but `$this` pseudo value is NOT a reference!! ( https://github.com/php/doc-en/pull/380 )

This PR works on PHP 7.4.x, 8.0.x.

### DIFF
--- a/phpdotnet/phd/Format.php
+++ b/phpdotnet/phd/Format.php
@@ -150,7 +150,7 @@ abstract class Format extends ObjectStorage
         $this->sqlite->query('SELECT example(docbook_id) FROM ids WHERE element=\'example\' OR element=\'informalexample\'');
     }
 
-    public function SQLiteIndex(&$context, $index, $id, $filename, $parent, $sdesc, $ldesc, $element, $previous, $next, $chunk) {
+    public function SQLiteIndex($context, $index, $id, $filename, $parent, $sdesc, $ldesc, $element, $previous, $next, $chunk) {
         $this->indexes[$id] = array(
             "docbook_id" => $id,
             "filename"   => $filename,
@@ -164,7 +164,7 @@ abstract class Format extends ObjectStorage
         );
     }
 
-    public function SQLiteChildren(&$context, $index, $id, $parent)
+    public function SQLiteChildren($context, $index, $id, $parent)
     {
         if (!isset($this->children[$parent])
             || !is_array($this->children[$parent])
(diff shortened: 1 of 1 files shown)


IMSoP:
The warning is nothing to do with `$this`, but I believe this fix is correct.

The warning comes when the [SQLite3 `createAggregate`](https://www.php.net/manual/en/sqlite3.createaggregate.php) method executes each of the callbacks passed here: https://github.com/php/phd/blob/master/phpdotnet/phd/Format.php#L138 The extension passes a value not a reference as the first parameter to those callbacks, triggering the warning.

Since none of the callbacks actually makes use of that first parameter, it should be perfectly fine to change signatures to accept it by value, as this PR does.

cmb69:
> Since none of the callbacks actually makes use of that first parameter, except for SQLiteFinal which returns it unchanged, it should be perfectly fine to change signatures to accept it by value, as this PR does.

Right, but that class might be extended, and we can't control that.

Also, we need to cater to https://github.com/php/phd/blob/a13d8b1edc4b1e3d9c04c6dfb9203445586f1cc4/composer.json#L21 (should probably have raised that long ago).

Girgias:
I think the first step is to raise the project dependency to 7.2 (which gives us *most* useful typing structures).

kamil-tekiela:
Do we actually know if anyone has created any extensions to this project? Did anyone ever try to override these methods? What would happen if they did so? Does the pass by reference do anything ever?

nikic:
I'd suggest putting a note in the README that this project has no API guarantees whatsoever and go wild :)

I don't think we (the PHP project) benefit from anything else.

cmb69:
Can we remove the package from PEAR and packagist?

salathe (approved):
Declaring these methods as taking the `$context` parameter by reference is a bug, as evidenced by the `E_WARNING`s.  If (and it's a big *if* !) anyone extends these methods in their own code (we don't in `phd`), their code has a bug and should be corrected.

TL;DR - IMO, don't sweat it, fix the signatures.

salathe:
> Can we remove the package ...

The following addresses the question of "can we remove", and not "should we remove".

> ... from PEAR ... 

The PEAR package is in the doc.php.net channel (rather than the usual pear.php.net channel), which is managed via the [php/web-doc](https://github.com/php/web-doc) Git repo.

> ... and packagist?

I don't see a way to remove the [php/phd package](https://packagist.org/packages/php/phd) from Packagist, but have marked it as "abandoned".  @Seldaek is there a self-service way to remove that package, or would someone from your side need to remove it?

Seldaek:
I removed the package from packagist.

salathe:
Thanks for the assistance, Jordi. 👍

cmb69 (approved):

PR #41, opened 2021-06-11
Remove whitespace before semicolons

kamil-tekiela:
The `in_array` is used to make it more compatible with https://github.com/php/phd/pull/38/files

This PR removes annoying and inconsistent whitespace before `;` in `<fieldsynopsis>` (properties). E.g. https://www.php.net/manual/en/class.mysqli-result.php

### DIFF
--- a/phpdotnet/phd/Render.php
+++ b/phpdotnet/phd/Render.php
@@ -160,7 +160,12 @@ class Render extends ObjectStorage
 
                 case \XMLReader::WHITESPACE: /* {{{ */
                 case \XMLReader::SIGNIFICANT_WHITESPACE:
-                            /* WS is always WS */
+                /* The following if is to skip whitespace before closing semicolon after property name */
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you, a nice improvement! However, this conflicts with PR #38. @kocsismate, what do you think: combine both PRs, or merge this and rebase yours?

Or has someone general objections regarding this patch?

kamil-tekiela:
As far as I understand @kocsismate only change methodsynopsis. My PR changes the fieldsynopsis. So, if my PR is merged then Mate will have to rebase and handle the merge, which is why I used `in_array` to make this easier

kocsismate:
I agree this is very nice an improvement, and I don't think anyone would miss the extra whitespace before the semicolon. I'm also ok to rebase if this PR gets merged.

cmb69 (approved):

PR #42, opened 2021-06-11
Fix PHP warnings

kamil-tekiela:
This fixes the annoying:
>[21:40:50 - E_WARNING             ] D:\projects\phd\phpdotnet\phd\Format.php:150
        phpdotnet\phd\Format::SQLiteFinal(): Argument #1 ($context) must be passed by reference, value given

I don't know if pass-by-ref was needed in previous PHP versions, but the current documentation says it's a normal parameter.

### DIFF
--- a/phpdotnet/phd/Format.php
+++ b/phpdotnet/phd/Format.php
@@ -150,7 +150,7 @@ abstract class Format extends ObjectStorage
         $this->sqlite->query('SELECT example(docbook_id) FROM ids WHERE element=\'example\' OR element=\'informalexample\'');
     }
 
-    public function SQLiteIndex(&$context, $index, $id, $filename, $parent, $sdesc, $ldesc, $element, $previous, $next, $chunk) {
+    public function SQLiteIndex($context, $index, $id, $filename, $parent, $sdesc, $ldesc, $element, $previous, $next, $chunk) {
         $this->indexes[$id] = array(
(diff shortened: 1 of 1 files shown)


cmb69:
> […], but the current documentation says it's a normal parameter.

Where are these docs?

Regardless of docs this change is a BC break, and needs to be handled as such; not sure what exactly that means, tough.

kamil-tekiela:
I meant this docs https://www.php.net/manual/en/sqlite3.createaggregate.php

Why would that be a breaking change? Because the function is a public function? I think we can make them private, SQLite3 shouldn't complain about that.

cmb69:
The `phpdotnet\phd\Format::SQLite*()` methods are supposed to be overrideable, I think. I also think that the signature in the docs is wrong, since they also state: "Your PHP function should accumulate the result and store it in the aggregation context."

kamil-tekiela:
>and store it in the aggregation context.
Which is then returned from the function. Anyway, if the manual is wrong why does PHP throw a warning due to invalid signature. 

I don't know how this project works. If there is a chance that someone overrides these methods then we should handle it more carefully, but I see no way to handle it differently right now. A bug is a bug and it needs to be fixed. If it breaks someone else's code then they need to fix that bug too.

kamil-tekiela:
Duplicate of #39

PR #44, opened 2021-06-18
Fix newline handling in case of constructors

kocsismate:


cmb69 (approved):
Thank you!

phpdotnet/phd/Render.php
```diff
@@ -71,7 +71,7 @@ public function execute(Reader $r) { /* {{{ */
                 }
 
                 $innerXml = null;
-                if ($r->name === "methodsynopsis" && $open) {
+                if (in_array($r->name, ["methodsynopsis", "constructorsynopsis", "destructorsynopsis"]) && $open) {
```
kamil-tekiela, line 74: If there's no reason to do otherwise, please always use strict checking for `in_array`

kocsismate: Yeah, I normally use the strict mode. I forgot it in this case, and didn't care to fix it as it wouldn't make any distinction here.

PR #47, opened 2021-07-17
Support rendering intersection types

cmb69:
As of PHP 8.1.0, intersection types are supported, so we already
prepare PhD for rendering respective types.  Since for now only pure
union and intersection types are supported, we don't handle complex
cases, although these are allowed by our DocBook DTD.

---

This requires https://github.com/php/doc-base/pull/37.

cc @Girgias

### DIFF
--- a/phpdotnet/phd/Package/PHP/XHTML.php
+++ b/phpdotnet/phd/Package/PHP/XHTML.php
@@ -155,8 +155,11 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
         "refsynopsisdiv"               => null,
     );
 
-    /** @var int|null Number of already formatted types in the current union type */
+    /** @var int|null Number of already formatted types in the current compound type */
     private $num_types = null;
+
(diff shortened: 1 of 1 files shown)


Girgias:
LGTM for a basic implementation.

However, depending on how compound types work it might need to be worked on again such that it is rendered as `(A|B)&C`/`(X&Y)|Z`/etc., but I would imagine one would need to check if there are nested type elements for that (and this can wait).

cmb69:
To be clear, mixed union and intersection types are not supported at all with this patch (nor are nested union or intersection types); at a minimum, we would need to implement a stack to store the information (`$type_separator` and `$num_types`).

Girgias:
> 
> 
> To be clear, mixed union and intersection types are not supported at all with this patch (nor are nested union or intersection types); at a minimum, we would need to implement a stack to store the information (`$type_separator` and `$num_types`).
> 
> PS: but YAGNI

That's what I thought :D but yeah will get round to that when we need to

PR #48, opened 2021-07-25
Fix formatting of the implements keyword

kocsismate:
While "extends" has a special formatting, "implements" is currently missing this. E.g. https://www.php.net/manual/en/class.appenditerator.php


cmb69 (approved):
Thank you! LGTM.

I wonder, though, why we use `<modifier>extends</modifier>` for `<ooclass>` but not `<modifier>implements</modifier>` for `<oointerface>`. Well, actually I understand that it is simpler to implement it that way, but it still is a bit strange. OTOH, changing this would be quite an issue regarding existing docs/translations.

kocsismate:
> OTOH, changing this would be quite an issue regarding existing docs/translations.

Yeah, initially I wanted to implement it via using `modifier`, but then changed my mind due to "BC" issues.

PR #50, opened 2021-07-25
Add support for displaying class/interface on class synopsis pages

kocsismate:
A few examples:

class

class-final

interface

class-extends

class-implements-one

class-implements-many

class-extends-implements


cmb69:
Using oointerface/interfacename for interfaces is certainly an improvement. I *presume* that this implementation is good as well, but I think we really should have CI with tests, and extend the testsuite.

kocsismate:
> but I think we really should have CI with tests, and extend the testsuite.

Yeah, I still agree with it, but that would require quite a few additional effort, for which I don't really have the bandwidth. :(

cmb69:
Seems we need to follow up with manual fixes: :)

Screenshot 2021-07-30 184631

PR #51, opened 2021-07-25
Improve the display of union and intersection types

kocsismate:
Simple nullable types are now displayed via using "?", while support for intersection types is reintroduced.

Examples for intersection types on the parameter and return type position (even though `X&null` is not valid):
Screenshot 2021-07-25 at 21 39 36

Examples for displaying "simple union types" via using "?":
Screenshot 2021-07-25 at 21 46 37


cmb69 (approved):
Very nice, thank you!

PR #57, opened 2021-10-19
Fix PHP 8.1 deprecations.

Crell:
What it says on the tin.

One issue is an API incompatibility with `SplObjectStorage`, which I am not fixing for now, just side-stepping.  Fixing it is going to be a larger issue I don't have bandwidth or context for at this point.


salathe:
Thanks @Crell ! Merged your changes, and made a few extras in 68363f64. 👍

PR #58, opened 2021-10-29
make objectstorage::attach() compatible with splobjectstorage::attach()

salathe:
PhD's `ObjectStorage::attach()` method does not respect the declared return type from its parent class (`SplObjectStorage`): the latter declares the return type as `void`.

This will issue an `E_DEPRECATED` as of PHP 8.1.0. It is currently mitigated via the `#[\ReturnTypeWillChange]` attribute (added in 6c1af76a).

This PR removes that attribute and instead aligns the return type declaration (to return `void`), with associated fixes to not return anything and not try to use a return value.

Since declaring a `void` return type is only allowed as of PHP 7.1.0, this PR raises the minimum required PHP version for phd to that version. Since PHP 7.1 is super ancient and even PHP 7.3 will become EOL in a few weeks (at the time of writing) I'm making the executive decision that requiring 7.1 is fine (though feedback on this point in particular is welcomed). 🙃

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ jobs:
   ubuntu:
     strategy:
       matrix:
-          version: ['5.6', '7.2', '7.4', '8.1']
+          version: ['7.1', '7.2', '7.4', '8.1']
     runs-on: ubuntu-latest
     steps:
       - name: Checkout PhD
--- a/README
+++ b/README
@@ -25,7 +25,7 @@ Installing the PhD renderer:
             for i in package_*.xml; do pear install $i; done
 
 Requirements:
-    - PHP 5.3
+    - PHP 7.1
     - DOM, libXML2, XMLReader and SQLite3.
 
 After installing PhD you can use the `phd` command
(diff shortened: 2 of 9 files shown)


salathe:
For a point of reference on the (change of) minimum version requirements, I kind of want to bump this all of the way up to 7.4 (since 7.3 is EOL in a few weeks) but the build servers (docs and rsync) are still on 7.3 😊, so stayed conservative with the change to 7.1.

kamil-tekiela:
I would be happy with raising it to PHP 7.4 or even PHP 8.0. Maybe we should consider upgrading the build system to be compatible with PHP 8.0. If we must support PHP 7.3 for the moment then maybe we can raise it to PHP 7.3 for the time being? PHP should be giving an example and keep up with the new versions. If it's ok for us to be still using PHP 7.1 how can we convince others to upgrade?

nikic (approved):
This is a reasonable start in any case, can always bump the version further later :)

cmb69 (approved):
Yes, fully agree with @nikic.

salathe:
Thanks all for the feedback and approvals. 🥇

PR #60, opened 2021-11-24
Remove curly brace array offset

iggyvolz:
Curly braces for array access were deprecated in PHP 7.4.  It looks like the extension this class uses (https://pecl.php.net/package/haru/1.0.4) isn't maintained and is no longer compatible with PHP - might be safe to remove it?


cmb69:
Yeah, the PDF rendered is unused (since 2012 or so).

kamil-tekiela:
We could merge this PR. There are more instances of this issue as well as numerous other issues. Maybe, the first step should be removing the PDF renderer?

cmb69:
+1 on removing the PDF renderer

haszi:
Closing this as there are no more curly brace array offsets in the codebase and the PDF renderer has been removed.

PR #52, opened 2021-07-29
Convert autoloading to PSR-4 and use composer's autoload

Girgias:
This requires to bump to PHP 8.0 so that 'Abstract' can be used in a namespace

This might give us better IDE support and better ways to setup testing...

I tried to test this by running it but I can't figure out which command I am meant to enter so I'm hitting issues with trying to render PDF which is expecting the Haru extension to be loaded.... which has been abandoned for ages.

For reference the command I entered is: `php render.php -d ../base/.manual.xml -P PHP`

PR #62, opened 2021-12-24
Use PSR-4 rather than PSR-0

iggyvolz:
https://www.php-fig.org/psr/psr-0/: 
> Deprecated - As of 2014-10-21 PSR-0 has been marked as deprecated. PSR-4 is now recommended as an alternative.

Used the PHPStorm refactoring tools as well as some manual labor to move all the classes to their correct location and use namespaces rather than underscores.  Also uses Composer's autoloader rather than the ad-hoc one.

Definitely needs some more testing to make sure it covers every possible class name access - there are some places where classes were accessed by strings but I _think_ I got all of those.

PDF renderer removed in the process, as per https://github.com/php/phd/pull/60#issuecomment-977993143.

Feel free to close if this is too ambitious or too much change all at once.


composer.json
```diff
@@ -21,14 +21,23 @@
         "php": ">=7.1.0",
         "ext-dom": "*",
         "ext-sqlite3": "*",
-        "ext-xmlreader": "*"
+        "ext-xmlreader": "*",
+        "ext-simplexml": "*"
```
iggyvolz, line 25: SimpleXML is used in phpdotnet\phd\Package\IDE\API so I added that here

src/Options/OptionsInterface.php
```diff
@@ -1,7 +1,7 @@
 <?php
-namespace phpdotnet\phd;
+namespace phpdotnet\phd\Options;
 
-interface Options_Interface {
+interface OptionsInterface {
```
iggyvolz, line 4: Name here should be "Interface" which is not a valid class name

src/Format/_Abstract/XHTML.php
```diff
@@ -1,7 +1,12 @@
 <?php
-namespace phpdotnet\phd;
+namespace phpdotnet\phd\Format\_Abstract;
```
iggyvolz, line 2: Abstract cannot be a namespace part - using _Abstract here

src/Package/IDE/API/_Function.php
```diff
@@ -21,7 +24,7 @@
  * @license  http://www.opensource.org/licenses/bsd-license.php BSD Style
  * @link     https://doc.php.net/phd/
  */
-class Package_IDE_API_Function
+class _Function
```
iggyvolz, line 27: Function is not a valid PHP class name

Girgias:
See #52

iggyvolz:
> See #52

Whoops not sure how I missed that, sorry for the noise.

PR #63, opened 2022-04-25
replaced <acronym> with <abbr> for HTML5.

mumumu:
In HTML5, `<acronym>` tag is classified as a non-conforming-features.

https://html.spec.whatwg.org/multipage/obsolete.html#non-conforming-features

### DIFF
--- a/phpdotnet/phd/Package/PHP/XHTML.php
+++ b/phpdotnet/phd/Package/PHP/XHTML.php
@@ -621,7 +621,7 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
     public function format_acronym_text($value, $tag) {
         $resolved = $this->acronymInfo($value);
         if ($resolved) {
-            return '<addr title="' .$resolved. '">' .$value. '</addr>';
+            return '<abbr title="' .$resolved. '">' .$value. '</abbr>';
         }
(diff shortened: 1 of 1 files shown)


phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -621,9 +621,9 @@ public function acronymInfo($acronym) {
     public function format_acronym_text($value, $tag) {
         $resolved = $this->acronymInfo($value);
         if ($resolved) {
-            return '<acronym title="' .$resolved. '">' .$value. '</acronym>';
+            return '<addr title="' .$resolved. '">' .$value. '</addr>';
```
cmb69, line 624: Typo:
```suggestion
            return '<abbr title="' .$resolved. '">' .$value. '</abbr>';
```

cmb69 (commented):
Ah, good catch! Do we also need to adjust [the CSS](https://github.com/php/web-php/tree/master/styles)?

mumumu:
> Do we also need to adjust [the CSS](https://github.com/php/web-php/tree/master/styles)?

Yes! I looked into web-php/styles, but could not found `acronym` selector.

https://github.com/php/web-php/search?q=acronym&type=code

cmb69 (approved):
Thank you!

PR #64, opened 2022-05-02
Add interfaces render

mallardduck:
This is related to:
- https://github.com/php/doc-base/pull/61
- https://github.com/php/doc-en/issues/1534

This renders to:
Screen Shot 2022-05-02 at 12 53 20 PM

### DIFF
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -393,14 +393,17 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
             'constructorsynopsis' => array(
                 /* DEFAULT */     false,
                 'classsynopsis' => 'format_classsynopsis_methodsynopsis_methodname_text',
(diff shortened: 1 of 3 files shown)


mallardduck:
I've read somewhere that the PDF output is deprecated and not used by PHP core. As such I've excluded the changes for that code route that would be necessary for this PR's goal to work with PDF output.

mallardduck:
I'm gonna close this PR out since @cmb69 pointed out in another PR that changing the DocBook (in doc-base) won't be an accepted option. So I closed out that attempt and PR. However, since the latest DocBook uses RELAX NG and RELAX NG is extensible it should be possible to accomplish using the officially documented method here: https://tdg.docbook.org/tdg/4.5/ch05.html

Note the caveat: 
> If You Change DocBook, It's Not DocBook Anymore!

So I guess the same idea would apply to PHP's current (older) DocBook. However before any "proper" extensions can be done it seems that updating to current Relax NG based DocBook would be needed first (whatever that looks like, it's beyond my knowledge).

That said, this PR won't be much help by the time any of that gets done (if ever). Especially since #69 aims to fix PDF output and this PR considered that deprecated and didn't address that.

PR #66, opened 2022-06-28
Fix: Stop invoking useless function

localheinz:
This pull request

- [x] stops invoking a useless function `manual_header()`

💁‍♂️ For reference, see https://github.com/php/web-php/pull/584.

### DIFF
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -204,7 +204,6 @@ $setup["toc_deprecated"] = $TOC_DEPRECATED;
 $setup["parents"] = $PARENTS;
 manual_setup($setup);
 
-manual_header();
 ?>
 ';
     }


kamil-tekiela:
I am not convinced. Can we not decouple PHD from WEB? Why is one calling functions from the other?

cmb69:
PhD doesn't call that function, but rather emits the function call to the manual pages.

@philip, do you have any concerns about this change; AFAIK, you're no longer using the latest versions of PhD, but I want to make that we're not breaking anything for you.

philip:
Thank you @cmb69 for remembering and checking in, I truly appreciate it.

We recently stopped importing PHP's MySQL connector documentation into the MySQL documentation so our use case can be ignored; it's probably safe to treat PhD as a php.net specific project.

salathe (approved):

cmb69:
Thanks for the confirmation, @philip! And thank you for the PR, @localheinz!

localheinz:
Thank you, @cmb69, @kamil-tekiela, @philip, and @salathe!

PR #67, opened 2022-06-29
Fix missing value issue

Crell:
I am not clear on how or why this happens, but sometimes for me index 3 is missing, which results in a missing index warning followed by a null-parameter deprecation message.  This fix addresses both.


cmb69:
> but sometimes for me index 3

From what I can tell, that would hint at a missing changelog description entry. If that is true, we probably should not work around that, but rather fix the manual.

cmb69:
I've removed that changelog (entry) now. Please double-check that the issue is resolved, @Crell.

Crell:
Yep, all clear.

PR #68, opened 2022-07-30
Add total number of functions and methods to the index

cmb69:
Apparently, users are interested in this number; there are constantly
new user notes which mention this number (a lot of these have been
deleted in the past).  This patch would display the total number of
functions and methods, so users won't need to calculate for themselves.

As it is now, i18n is completely missing.  We would need to define an
XML entity in language-snippets.ent for this.

### DIFF
--- a/phpdotnet/phd/PI/PHPDOCHandler.php
+++ b/phpdotnet/phd/PI/PHPDOCHandler.php
@@ -60,9 +60,9 @@ class PI_PHPDOCHandler extends PIHandler {
                             $longdesc = $info[$filename][1] ? " - {$info[$filename][0]}" : "";
                             $ret .= '<li><a href="'.$filename. '" class="index">' .$data. '</a>' . $longdesc . '</li>'."\n";
                         }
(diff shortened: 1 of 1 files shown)


Girgias (approved):
It is really bizarre why people care about it, but this LGTM and we can always add internationalization later.

cmb69:
Just saw @salathe's downvote. Is it about the missing internationalization, or some other detail, or presenting the total number at all?

salathe:
> Just saw @salathe's downvote. Is it about the missing internationalization, or some other detail, or presenting the total number at all?

Mostly the latter: presenting the total number at all.  

The number is not the count "things" (the page title says functions and methods, but other things are listed in that index) in PHP at a point in time, nor in any specific version of PHP, nor in even PHP at all (each of which _might_ be useful numbers). It is just however many "things" are in the docs at a point in time; a number that I don't consider useful.   The user notes that were providing a count (most of them just updating the number from previous years' notes) have been deleted.

My vote, for whatever it is worth, would be to not bother with this change.

cmb69:
Fair enough. :)

Issue #71, opened 2022-10-30
In Firefox, the code from the documentation is copied with Non-breaking spaces (nbsp)


deevroman:
Starting with Firefox 103, it now does not replace nbsp with a normal space. https://bugzilla.mozilla.org/show_bug.cgi?id=359303

Try copying some code from the documentation and pasting it into a text file. (for example from here https://www.php.net/manual/en/function.preg-replace.php)

<img width="614" alt="Снимок экрана 2022-10-30 в 19 15 36" src="https://user-images.githubusercontent.com/25708359/198889411-2c4ca826-40cc-4126-9b44-e8c785b4a601.png">

ianthedev:
> I think we should follow suite and get rid of the non-breaking spaces, and instead prevent space-folding by using `white-space: pre` or such.

Agreed. I'm a Firefox user as well and encountered this issue yesterday.

> Anyhow, this is actually a PhD doc issue.

Should the discussion be moved to PhD doc or continue here? I'm seeing no one here has been assigned to this issue.

damianwadley:
> Should the discussion be moved to PhD doc or continue here?

"Here" _is_ phd.

ianthedev:
> > Should the discussion be moved to PhD doc or continue here?
> 
> "Here" _is_ phd.

I see. Didn't realize "phd" is the same as "PhD doc".

cmb69:
> I'm seeing no one here has been assigned to this issue.

Given that this is an open-source project, you could have submitted a pull request. :)

ianthedev:
> > I'm seeing no one here has been assigned to this issue.
> 
> Given that this is an open-source project, you could have submitted a pull request. :)

I see. Didn't know this issue can be solved that way. Besides, submitting a pull request is current beyond my ability. In fact, I practiced that many years ago by using GitHub Desktop and have long since forgotten the steps. But thanks for letting me know a solution.

PR #72, opened 2022-12-23
Fix GH-71: Code examples contain non-breaking spaces

cmb69:
Some browsers (notably Firefox ≥ 107) copy these NBSPs verbatim, what causes issues when pasting into text editors/IDE.  Since these NBSPs are generated by `highlight_string()` and there is no option to produce regular spaces instead, we replace them manually.

To preserve the desired code layout, this requires something like <https://github.com/php/web-php/pull/744>.

--- 

Note that this change only affects PHP code samples; there are other cases where PhD renders NBSPs, but these can be addressed when and if needed.

### DIFF
--- a/phpdotnet/phd/Highlighter.php
+++ b/phpdotnet/phd/Highlighter.php
@@ -63,7 +63,7 @@ class Highlighter
 
         if ($role == 'php') {
             try {
-                return highlight_string($text, 1);
+                return str_replace('&nbsp;', ' ', highlight_string($text, 1));
             } catch (\ParseException $e) {
                 v("Parse error while highlighting PHP code: %s\nText: %s", (string) $e, $text, E_USER_WARNING);
 


kamil-tekiela:
What about newlines? `highlight_string()` produces a lot of them in non-necessarily valid places.

cmb69:
> What about newlines?

You mean wrt. `white-space:pre`? I haven't noticed any issues, but I only checked a couple of examples.

The obvious alternative would be to avoid `highlight_string()`, and to use a client side script to do the highlighting (I wouldn't want to revive the probably broken geshi support in PhD). I'm not against that, but it would be a more severe change. Shall I investigate about that?

kamil-tekiela:
Don't spend too much time on it. I'd much rather see `highlight_string` patched up. Your PHD fix seems ok. I tested it already a week ago and it should work, apart from the extra newlines. I think we can merge this fix for now.

TimWolla:
> Shall I investigate about that?

FWIW: At work we use Prism and I'm pretty happy with it: https://prismjs.com/

ianthedev:
I'm using Firefox 108.0.1 (64-bit) on Windows 7, and copying the [example code of `mb_strimwidth`](https://www.php.net/manual/en/function.mb-strimwidth.php#refsect1-function.mb-strimwidth-examples) still causes the same issue.

cmb69:
@ianthedev, that patch will take a couple of hours to be rolled out.

PR #75, opened 2023-06-29
Add rendering logic for interfaces

mallardduck:
- Adds cchunk variable for tracking if a classsynopsis is actually an interface
- Render extends instead of implements for interfaces
- Render interface instead of class

After this is merged in, then this PR will render DS as interfaces now: https://github.com/php/doc-en/pull/2549

### DIFF
--- a/phpdotnet/phd/Package/Generic/PDF.php
+++ b/phpdotnet/phd/Package/Generic/PDF.php
@@ -277,6 +277,7 @@ abstract class Package_Generic_PDF extends Format_Abstract_PDF {
         "classsynopsis"         => array(
             "close"             => false,
             "classname"         => false,
+            "interface"         => false, // bool: true when in interface
         ),
         "classsynopsisinfo"     => array(
             "implements"        => false,
@@ -932,7 +933,12 @@ abstract class Package_Generic_PDF extends Format_Abstract_PDF {
         if ($open) {
             if ($this->cchunk["classsynopsisinfo"]["implements"] === false) {
                 $this->cchunk["classsynopsisinfo"]["implements"] = true;
-                $this->pdfDoc->appendText(" implements");
+                if ($this->cchunk["classsynopsis"]["interface"]) {
+                    $this->pdfDoc->appendText(" extends");
+                } else {
+                    $this->pdfDoc->appendText(" implements");
+                }
+
                 return '';
             }
             $this->pdfDoc->appendText(",");
@@ -943,6 +949,13 @@ abstract class Package_Generic_PDF extends Format_Abstract_PDF {
 
     public function format_classsynopsis($open, $name, $attrs, $props) {
         if ($open) {
+            if (
+                isset($attrs[Reader::XMLNS_DOCBOOK]["class"]) &&
+                $attrs[Reader::XMLNS_DOCBOOK]["class"] == 'interface'
+            ) {
+                $this->cchunk["classsynopsis"]["interface"] = true;
+            }
+
             return $this->format_para($open, $name, $attrs, $props);
         }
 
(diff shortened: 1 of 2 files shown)


mallardduck (commented):
adjust styles to match better

heiglandreas:
What irritates me TBH is that it looks like the possibilities are already there.

See this rendering of `Iterator` available at https://www.php.net/manual/en/class.iterator.php

image

On the other side there's this rendering of the `Ds\Collection` interface available at https://www.php.net/manual/en/class.ds-collection.php

image

The former is rendered from this information:

https://github.com/php/doc-en/blob/dd9775e98b7368f4d04847f343250c9627bb8da8/language/predefined/iterator.xml#L20-L47

whereas the later is rendered from this information: https://github.com/php/doc-en/blob/dd9775e98b7368f4d04847f343250c9627bb8da8/reference/ds/ds.collection.xml#L22-L46

From what I see the main difference is this:

```
    <classsynopsisinfo>
     <oointerface>
      <interfacename>Iterator</interfacename>
     </oointerface>


     <ooclass>
      <modifier>extends</modifier>
      <classname>Traversable</classname>
     </ooclass>
    </classsynopsisinfo>
```
compared to
```
    <classsynopsisinfo>


    <ooclass><classname>Ds\Collection</classname></ooclass>


    <oointerface><interfacename>Countable</interfacename></oointerface>
    <oointerface><interfacename>IteratorAggregate</interfacename></oointerface>
    <oointerface><interfacename>JsonSerializable</interfacename></oointerface>


    </classsynopsisinfo>
```

So it looks like everything is actually already on board for what you want.

Or do I miss something?

mallardduck:
This won't work because that DTD is designed for true classes - which can only ever extend one thing where as interfaces can extend multiple other interfaces. So in our situation with DS - trying a similar structure we would use:

```
     <ooclass>
      <modifier>extends</modifier>
      <classname>Countable</classname>
      <classname>IteratorAggregate</classname>
      <classname>JsonSerializable</classname>
     </ooclass>
```

However this produces an error:

```
The document didn't validate, here are the errors I got:
(If this isn't enough information, try again with --enable-xml-details)
 Element ooclass content does not follow the DTD, expecting ((package |
 modifier)* , classname), got (modifier classname classname classname )
```

heiglandreas:
Thanks for the explanation. So I indeed missed something 😁

phpdotnet/phd/Package/Generic/PDF.php
```diff
@@ -943,6 +949,13 @@ public function format_classsynopsisinfo_oointerface($open, $name, $attrs) {
 
     public function format_classsynopsis($open, $name, $attrs, $props) {
         if ($open) {
+            if (
+                isset($attrs[Reader::XMLNS_DOCBOOK]["class"]) &&
```
heiglandreas, line 953: I assume that the `class`-attribute is declared in the DTD. Otherwise it might be a good idea to rename that perhaps to "type" as that is describing better what we are referencing here.

heiglandreas (commented):
Looks good from my point of view.

phpdotnet/phd/Package/Generic/PDF.php
mallardduck: Yeah - this is correct. The class attribute is already defined with valid values of either class or interface.

https://tdg.docbook.org/tdg/4.5/classsynopsis.html

Girgias: More up to date version of the DTD we are aiming to use: https://tdg.docbook.org/tdg/5.1/classsynopsis.html

mallardduck: Thanks for more current link @Girgias,  looks like it still supports the same attribute used in this PR tho. I just happened to relink to the original link @cmb69 provided the first time this method was suggested in a previous PR.

kocsismate:
I also added support for this in gen_stub.php: https://github.com/php/php-src/commit/3c6590a3911cdc5841f56edef95a2eac7fe1da31

PR #79, opened 2023-08-08
Drop `\n` when highlighting

TimWolla:
In a second step the CSS will need to be adjusted.

------------

Actual newlines will be replaced by `<br />` during highlighting and the `\n` in the boilerplate conflicts with `white-space: pre;`.

see php/doc-en#2648

### DIFF
--- a/phpdotnet/phd/Highlighter.php
+++ b/phpdotnet/phd/Highlighter.php
@@ -63,7 +63,11 @@ class Highlighter
 
         if ($role == 'php') {
             try {
-                return str_replace('&nbsp;', ' ', highlight_string($text, 1));
+                $highlighted = highlight_string($text, 1);
+                $highlighted = str_replace('&nbsp;', ' ', $highlighted);
+                $highlighted = str_replace("\n", '', $highlighted);
+
+                return $highlighted;
             } catch (\ParseException $e) {
                 v("Parse error while highlighting PHP code: %s\nText: %s", (string) $e, $text, E_USER_WARNING);
 
--- a/phpdotnet/phd/Highlighter.php
+++ b/phpdotnet/phd/Highlighter.php
@@ -63,11 +63,10 @@ class Highlighter
 
         if ($role == 'php') {
             try {
-                $highlighted = highlight_string($text, 1);
-                $highlighted = str_replace('&nbsp;', ' ', $highlighted);
-                $highlighted = str_replace("\n", '', $highlighted);
-
-                return $highlighted;
+                return strtr(highlight_string($text, 1), [
+                    '&nbsp;' => ' ',
+                    "\n" => '',
+                ]);
             } catch (\ParseException $e) {
                 v("Parse error while highlighting PHP code: %s\nText: %s", (string) $e, $text, E_USER_WARNING);
 


phpdotnet/phd/Highlighter.php
```diff
@@ -63,7 +63,11 @@ public function highlight($text, $role, $format)
 
         if ($role == 'php') {
             try {
-                return str_replace('&nbsp;', ' ', highlight_string($text, 1));
+                $highlighted = highlight_string($text, 1);
+                $highlighted = str_replace('&nbsp;', ' ', $highlighted);
+                $highlighted = str_replace("\n", '', $highlighted);
+
+                return $highlighted;
```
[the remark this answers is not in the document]
TimWolla: Thanks, I initially had some more complex `preg_replace()` call there. I've opted to use `strtr()`, as I find the key->value array more readable than two separate arrays.

kamil-tekiela:
For an easier understanding, it would be nice if you could provide examples. How exactly does it interfere? Why do we now need to remove the `\n`?

TimWolla:
To properly include the original indentation of stuff output of `<?php … ?>` tags (i.e. embedded HTML), the existing:

```
.phpcode code span span { white-space: pre; }
```

formatting is insufficient, because the selector doesn't match.

However putting `white-space: pre` onto `.phpcode code` will cause bogus formatting, because of the included newlines (there is one at the beginning right in front of the code and one at the end):

image

Without the newlines:

image

kamil-tekiela:
Honestly, we should probably stop using `highlight_string()`. It breaks HTML standard in multiple ways and the resulting formatting is questionable. This is just one of many issues that we had on php.net. 
We abuse the `<code>` element which is meant to be an inline element. The proper solution to the current issue would be to wrap the `<code>` element within `<pre>` as is recommended. But as you noticed, we have newlines in weird spots within the highlighted string. 

We could fix the function within php-src because we technically are allowed to by the [PHP documentation](https://www.php.net/manual/en/function.highlight-string.php#refsect1-function.highlight-string-notes):

> The HTML markup generated is subject to change.

I think it would be the best fix, but we would need to wait for the next minor PHP release and then upgrade the environment for php.net. 

The fix would be easy:
- Instead of `<be>` output `\n`
- Instead of `&nbsp;` output ` `
- Instead of `<code>` use `<pre>`
- Instead of `<span>` use `<code>`
- Remove the first enclosing `<code>` (old `<span>`)
- Remove the enclosing newlines

No more breaking HTML standard, no more adjustments of the output, and no more CSS styles on php.net. 

For the time being, I guess we could accept your PR.

phpdotnet/phd/Highlighter.php
```diff
@@ -63,7 +63,10 @@ public function highlight($text, $role, $format)
 
         if ($role == 'php') {
             try {
-                return str_replace('&nbsp;', ' ', highlight_string($text, 1));
```
Girgias, line 66: ``str_replace()`` can also take an array of values to replace and an array of what those values should be, why not just use that?

But the issue seems to be more that ``\n`` tags before and after are being converted to ``<br>`` so wouldn't it make more sense to just ``ltrim()`` new lines of the ``$text`` variable?

TimWolla: > `str_replace()` can also take an array of values to replace and an array of what those values should be, why not just use that?

see https://github.com/php/phd/pull/79#discussion_r1287289455

> But the issue seems to be more that \n tags before and after are being converted to <br> so wouldn't it make more sense to just ltrim() new lines of the $text variable?

This wouldn't work, because the extra newlines are inserted by `highlight_string()`, namely here:

https://github.com/php/php-src/blob/f8b42da3a4635aa2ace3d4dbfdbae6835fb43d7f/Zend/zend_highlight.c#L92

and here:

https://github.com/php/php-src/blob/f8b42da3a4635aa2ace3d4dbfdbae6835fb43d7f/Zend/zend_highlight.c#L165-L167

Girgias (approved):

PR #81, opened 2023-08-11
Detect and render `<emphasis role="strong|bold">` usage in manual

alfsb:
`<emphasis role="strong">` and `<emphasis role="bold">` are used in manual (occurs ~600 times), but are rendered as same as `<emphasis>`.

This PR detects and render these cases as `<strong>` instead of `<em>`.

For example, in https://www.php.net/manual/en/features.commandline.options.php it's change from
image
to
image

and in https://www.php.net/manual/en/mysqli.quickstart.dual-interface.php it's change from
image
to
image

### DIFF
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -80,7 +80,7 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
         ),
         'constructorsynopsis'   => 'format_methodsynopsis',
         'destructorsynopsis'    => 'format_methodsynopsis',
-        'emphasis'              => 'em',
+        'emphasis'              => 'format_emphasis',
         'enumname'              => 'span',
         'entry'                 => array (
             /* DEFAULT */          'format_entry',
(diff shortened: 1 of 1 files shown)


phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -1173,6 +1173,21 @@ public function format_classsynopsis_methodsynopsis_methodname_text($value, $tag
...
+        if (isset($attrs[Reader::XMLNS_DOCBOOK]['role'])) {
+            $role = $attrs[Reader::XMLNS_DOCBOOK]['role'];
+            if ( $role == "strong" || $role == "bold" )
+                $name = "strong";
+        }
+
+        if ($open)
+            return "<{$name}>";
+        else
+            return "</{$name}>";
```
Girgias, line 1189: CS: nit
```suggestion
        if ($open) {
            return "<{$name}>";
        } else {
            return "</{$name}>";
        }
```

phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -1173,6 +1173,21 @@ public function format_classsynopsis_methodsynopsis_methodname_text($value, $tag
...
     }
 
+    public function format_emphasis($open, $name, $attrs)
+    {
+        $name = "em";
+
+        if (isset($attrs[Reader::XMLNS_DOCBOOK]['role'])) {
+            $role = $attrs[Reader::XMLNS_DOCBOOK]['role'];
+            if ( $role == "strong" || $role == "bold" )
+                $name = "strong";
```
Girgias, line 1183:
```suggestion
            if ( $role == "strong" || $role == "bold" ) {
                $name = "strong";
            }
```

Girgias (approved):

PR #86, opened 2023-11-01
Fix missing version info

sy-records:
Fix php/web-php#818
Fix php/doc-en#2892


Girgias (approved):
I'm not sure why it doesn't work any more, but if this fixes it then sure.

PR #88, opened 2023-11-10
Fix constants link broken

### LINKED ISSUE #87: Broken links for constants in table rows
Hi,

The links to constants in https://www.php.net/manual/en/class.phar.php are broken (e.g., `https://www.php.net/manual/en/.php#phar.constants.sha256` )

While they point to the correct `id`s it seems the URLs for Ids inside `table.row` are not being generated correctly.

```
en/reference/phar/Phar.xml
121:     <varname linkend="phar.constants.sha256">Phar::SHA256</varname>

en/reference/phar/constants.xml
117:     <row xml:id="phar.constants.sha256">
```

While the code lies in the doc-en repository, I suppose this is a URL generation issue with phd (?)

sy-records:
Fix #87 

image

btw, there seems to be a problem with all the `<row xml:id` links.

e,g. https://www.php.net/manual/en/tokens.php

The `__TRAIT__` link is the current page and is correct after the fix.

image

--- 

Sorry, I committed to master by mistake. 😅 934be12873ce67e9db19586919afb2a45d791c91

### DIFF
--- a/phpdotnet/phd/Index.php
+++ b/phpdotnet/phd/Index.php
@@ -63,9 +63,9 @@ class Index extends Format
     'titleabbrev'           => 'format_sdesc',
     'example'               => 'format_example',
     'refsect1'              => 'format_refsect1',
-    "row"                   => array(
+    'tbody'                 => array(
         /* DEFAULT */          null,
-        "tbody"             => 'format_row',
+        'row'               => 'format_row',
     ),
     'entry'                 => array(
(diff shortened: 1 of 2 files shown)


Girgias (approved):
This looks correct to me, but I'm confused why we don't just index on all rows?

sy-records:
yep, I'm not sure why it wasn't there before, but now it is. 
image

PR #90, opened 2024-01-29
Add table formatting for constant lists

haszi:
@Girgias This is an attempt to implement formatting constant lists as tables. While most of it works fine, the `format_variablelist()` function has an issue  I'm hoping you might be able to help me with. Namely that I'm inserting a header row as a string here which includes the hardcoded strings `Constants` and `Description`. Is there a way to access the language specific strings of the appropriate entities (`&Constants;` and `&Description;`).


Girgias:
There is language data in ``phpdotnet/phd/data/langs`` but I'm not sure how this is accessed :/ let me have a look

Girgias:
Seems doing ``$this->autogen($key, $lang)`` fetches the relevant entry if it exists?

phpdotnet/phd/Format.php
```diff
@@ -395,6 +395,10 @@ final public static function autogen($text, $lang = null) {
 
         $filename = Config::lang_dir() . $lang . ".ini";
 
+        if (!file_exists($filename) && strncmp(basename($filename), 'doc-', 4) === 0) {
```
Girgias, line 398:
```suggestion
        if (!file_exists($filename) && strncmp(basename($filename), 'doc-', 4) === 0) {
```
Why not use ``str_starts_with()`` ?

phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -1476,16 +1476,27 @@ public function format_step($open, $name, $attrs) {
...
+                $headerStr = "\n<tr>\n<th>" . $this->autogen('Constants', $props['lang']) . "</th>\n<th>" . $this->autogen('Description', $props['lang']) . "</th>\n</tr>";
             } else {
-                return '<dl>';
+                $tagName = 'dl';
             }
+            return '<' . $tagName . $idStr . $classStr . '>' . $headerStr;
         }
-        return "</dl>\n";
+        $tagName = ($this->role === 'constant_list') ? 'table' : 'dl';
+        $this->role = false;
```
Girgias, line 1498: TIL there is a `role` property

Girgias (commented):
Haven't tried this locally yet, but this looks good :) would probably also allow use to automatically generate links for ``<constant>`` tags when all of the roles are set and so there is a way to index the constant definitions :)

I've pinged translators so they can have a look at the lang data.

phpdotnet/phd/Format.php
Girgias: Yeah it used to be that we cannot use PHP 8 because the doc server was running 7.3, but this is not the case anymore and I used it in https://github.com/php/phd/commit/365cd394f1b32a4b9583043ca10f8a2a9d07ba19#diff-c372d6edb12a842ef8a20b14089597d81f1b11f5e0fb7f98f9f305528b133ef5R892 :)

Probably a cleanup to do then :)

Girgias (commented):
Just had a look at it locally, and looks good :)

I do wonder maybe if we should move the type information into another ``term`` tag for it to be rendered as a column too, but that's going to require modifying a lot of doc pages :/

Girgias:
I think having the changelog (for addition/removal) can just be in the description. A whole column for it is wasteful IMHO.

Also I realised, considering the rendering only works when the markup is changed to add the role attribute, changing the markup at the same time to move the type info into a ``<term>`` tag should also be fine.

nielsdos:
I'll have a better look at home, but from https://www.w3.org/TR/xinclude-11/#attribute-copying I read:

> Also, while attributes in xml: namespace are not copied, it is possible to use the set-xml-id attribute to change xml:id attributes on XIncluded elements.

nielsdos:
> I'll have a better look at home, but from https://www.w3.org/TR/xinclude-11/#attribute-copying I read:
> 
> > Also, while attributes in xml: namespace are not copied, it is possible to use the set-xml-id attribute to change xml:id attributes on XIncluded elements.

But looks like libxml2 implements the 2003 spec of XInclude, not the newer XInclude version 1.1 that introduced the set-xml-id attribute. In that case I don't see a workaround except applying custom processing or XSLT. Or if you can solve it by including the children instead of the `<constant>` element itself that may work too depending on your use-case.

haszi:
@Girgias  please don't add the `constant_list` role to any `<variabelist>`s yet as nested `role`s will result in corrupted tables.

I was originally thinking about refactoring the `role` property to an array so that nested roles could be used for custom rendering but I'm already working on this to fix the bug/limitation on nested roles. 
I noticed this issue while moving all translations' cURL constants to their constants pages when one of the descriptions included a `<literal>` tag that the renderer added a `code` role to. And as the `role` property belongs to the format (not to an element) object the parent element's `constant_list` role was overwritten.

Is there a test suite for Phd somewhere I could test my changes on? I've found a handful of tests in the Phd directory but there seems to be way more functionality to Phd than this.

Girgias:
> @Girgias please don't add the `constant_list` role to any `<variabelist>`s yet as nested `role`s will result in corrupted tables.
> 
> I was originally thinking about refactoring the `role` property to an array so that nested roles could be used for custom rendering but I'm already working on this to fix the bug/limitation on nested roles. I noticed this issue while moving all translations' cURL constants to their constants pages when one of the descriptions included a `<literal>` tag that the renderer added a `code` role to. And as the `role` property belongs to the format (not to an element) object the parent element's `constant_list` role was overwritten.
> 
> Is there a test suite for Phd somewhere I could test my changes on? I've found a handful of tests in the Phd directory but there seems to be way more functionality to Phd than this.

I wasn't planning on adding it yet, as I'm leaving it to be your project in some way.

I saw you found the test suite already, which IIRC uses the php-src run-tests.php script

PR #94, opened 2024-02-11
Add links to type tags

haszi:
Add links to all `<type>`s in `<methodsynopsis>` and `<constructorsynopsis>` tags. Based on #33.

Includes the following>
 - refactored parameters to keep tracking of types the same way return types did
 - extracted functionality to handle simple nullable types and type `<span>` tags into its own method and used it for parameters too
 - added all types to the type link generating method
 - added link to the existing `void` rendering method
 - added tests for `<methodsynopsis>` and `<constructorsynopsis>` parameters, and `<methodsynopsis>` return types
 - updated an existing test that included types


haszi:
Note:  there are no tests for intersection types (neither on their own nor combined with union types) as I don't know how to write the appropriate xml and couldn't find any examples of it in the English docs.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -523,29 +510,109 @@ public function format_methodsynopsis($open, $name, $attrs, $props) {
...
 
-        return self::format_type_text($type, $tagname);
+    public function format_type_methodparam($open, $tag, $attrs) {
+        if ($open) {
+            if (isset($attrs[Reader::XMLNS_DOCBOOK]["class"])) {
+                if ($attrs[Reader::XMLNS_DOCBOOK]["class"] === "union") {
+                    $this->cchunk["methodparam"]["type_separator"] = "|";
+                } else {
+                    $this->cchunk["methodparam"]["type_separator"] = "&";
+                }
```
Girgias, line 595: I think using a `match` here and check that the `class` attribute is either `"union"` or `"intersection"` is better, this might. As the "default" should more be union types considering we have DNF types (which I don't think I remember how I was going about markup.

Also the `&` might need to be `&amp;` ? Depending on if the transformation results in XML or not.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -555,37 +622,55 @@ public function format_type_text($type, $tagname) {
...
+        case "false":
+        case "true":
+            $href = "language.types.value";
+            break;
+        case "null":
+            if ($this->simple_nullable) {
+                return "";
+            }
         case "boolean":
         case "integer":
```
Girgias, line 637: Those, plus the `"double"` case should warn IMHO, as those are an alternative spelling for the proper built-in types that should not exist in the documentation any more. But can be a follow-up PR.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -523,29 +510,109 @@ public function format_methodsynopsis($open, $name, $attrs, $props) {
...
+    }
+
+    public function format_type_methodparam_text($type, $tagname) {
+        $this->cchunk["methodparam"]["paramtypes"][] = $type;
+
+        return "";
     }
 
     public function format_type_text($type, $tagname) {
+        $type = trim($type);
```
Girgias, line 609: I'm not super sure we should do this, maybe add a warning? Because the sources shouldn't be having a whitespace ideally inside a `<type>` tag.

Girgias (commented):
Looks good just some questions/remarks

phpdotnet/phd/Package/PHP/XHTML.php
haszi: Right, using match with the exact words "union" or "intersection" is better here. It will also help with catching incorrect `class` values.

Good catch on the `&amp;`. I guess this has not been used anywhere yet as it was `&` all along. :-)

phpdotnet/phd/Package/PHP/XHTML.php
haszi: I was thinking about this and other warnings. There's a whole range of formatting and logic errors we could catch in Phd from whitespace in all sorts of tags, incompatible type combinations, etc.
If this was implemented as some sort of option that Phd could be called from the command line with, this whole doc checker could be added as an additional step to the documentations' pull request workflow. Maybe?

phpdotnet/phd/Package/PHP/XHTML.php
haszi: See my longer comment on warnings above. :-)

haszi: I'll remove this `trim()` though and the test case from the test files.

phpdotnet/phd/Package/PHP/XHTML.php
Girgias: I think this makes sense, but I was certain we already had support for warnings in PhD as I had seen them :thinking: 

Maybe if it's a new render that runs fast and doesn't need indexing, having it run on CI would make sense. Last time it came up there were concerns about the CI taking too long, or running too frequently as a PR might be update often in quick successions

tests/php/data/type_rendering_constructorsynopsis_parameters-and-return-type.xml
```diff
@@ -0,0 +1,78 @@
...
+  </constructorsynopsis>
+ </section>
+
+ <section>
+  <para>8. Constructor with union type parameter</para>
+  <constructorsynopsis>
+   <modifier>final</modifier> <modifier>public</modifier> <methodname>ClassName::__construct</methodname>
+   <methodparam><type class="union"><type>iterable</type><type>resource</type><type>callable</type><type>null</type></type><parameter>option</parameter></methodparam>
+  </constructorsynopsis>
+ </section>
```
Girgias, line 65: I think the "logical" way to have an intersection type in markup would be:
```xml
<type class="intersection"><type>Countable</type><type>Traversable</type></type>
```
and a DNF type would be:
```xml
<type class="union"><type class="intersection"><type>Countable</type><type>Traversable</type></type><type>DOMAttr</type></type>
```

But the rendering with the parenthesis for DNF types might be tricky and could be added later, intersection types tho I would imagine would "just" work.

Girgias (commented):
Final late remark :)

phpdotnet/phd/Package/PHP/XHTML.php
haszi: Oh, yeah. The error handler is right there in `functions.php`, a file included in everything Phd which I also never looked inside. :-)

So this format would be called with the `--noindex` option, it would have to override the `createLink()`, `update()` and `appendData()` methods at a minimum, and it's validation method(s) could return `null` and would raise warnings on validation errors. Sounds fairly simple, so it's probably far from it. :-)

tests/php/data/type_rendering_constructorsynopsis_parameters-and-return-type.xml
Girgias: Correct, there are _good_ implementation reasons as to why I forced it to be never more nested than `A|(B&C)` for the moment. :)

phpdotnet/phd/Package/PHP/XHTML.php
Girgias: As everything with PhD :D it looks simple and it isn't :D

tests/php/data/type_rendering_constructorsynopsis_parameters-and-return-type.xml
haszi: For the moment? :-(

What would be the correct `xml` output for your DNF example? Is the below correct or am I missing a `<span class="type">` and a `</span>` just inside the parentheses to wrap the intersection type? 
```xml
<span class="type">(<span class="type">Countable</span>&amp;<span class="type">Traversable</span>)|<span class="type">DOMAttr</span></span>
```

haszi:
I changed all `type_separator` properties to arrays and added a `type_separator_stack` property. The former holds the separator following the type with the corresponding index number in the `types` array, and the latter keeps track of the stack of separators during processing of tags/text and is used to fill the former.

tests/php/data/type_rendering_constructorsynopsis_parameters-and-return-type.xml
Girgias: We *might* need to support generalized type nesting for php-src for type aliases, but I'm not sure that's really something to bother about for the docs.

Looking at my docbook render, I settled on (added some line breaks):
```html
<span class="type">
    (<span class="type">Countable</span>&amp;<span class="type">Traversable</span>)|
    <span class="type">DOMAttr</span>
</span>
```

Girgias (approved):
Looks good to me :)

Sorry for the delay was at PHP UK

haszi:
> Looks good to me :)
> 
> Sorry for the delay was at PHP UK

I appreciate you taking the time to review these PR at all. :-) Especially that you made it very clear that Phd is not too close to your heart, to put it mildly. :-)

PR #95, opened 2024-02-13
Move whitespace formatting to generic XHTML format

haszi:
Move whitespace formatting from `Render.php` to the generic XHTML format and add the option for each format to control their own whitespace formatting. Add a test to ensure XHTML formatting is correct.

The current way of handling this in `Render.php` not only applies to all formats (XHTML, PDF, Manpage, etc.), it also does not allow formats to control the output of significant whitespace. With this PR, formats can implement the `format_whitespace()` method and override default whitespace output.

This PR also moves the existing conditional whitespace formatting to the generic XHTML format as I assumed this was specifically added for rendering the documentation in XHTML format.

Note:
this also corrects a minor inconvenience in formatting where whitespace after the very first `<methodsynopsis>`, `<constructorsynopsis>` or `<destructorsynopsis>` is printed. This was due to a non-existent `$elementStack[$currentDepth]` in `in_array($elementStack[$currentDepth] ?? "", ["methodname", "methodparam", "type", "void"], true)`.

Because of this fix, existing tests will need a minor update if this PR is merged which I will take care of.


haszi:
A note relating to the `TODO` in the generic XHTML format:

whitespace in `<type>` elements is not removed because XMLReader returns it as part of the `<type>`'s text. I'm planning on removing the lines that don't work (right under the `TODO`) and if there are no objections, add trim the text in the method handling the `<type>` elements' text.

tests/xhtml/data/whitespace_formatting_001.xml
```diff
@@ -0,0 +1,59 @@
...
+   <oointerface>
+    <interfacename>SecondInterface</interfacename>
+   </oointerface>
+   <oointerface>
+    <interfacename>ThirdInterface</interfacename>
+   </oointerface>
+  </classsynopsisinfo>
+ </section>
+
+</chapter>
```
Girgias, line 59: NIT EOL before EOF

tests/xhtml/whitespace_formatting_001.phpt
```diff
@@ -0,0 +1,76 @@
...
+  <p class="para">5. Implements list with whitespace</p>
+  <div class="classsynopsisinfo">
+   <span class="ooclass">
+    <span class="modifier">class</span> SomeClass
+   </span>
+   <span class="oointerface"><span class="modifier">implements</span> 
+     FirstInterface</span><span class="oointerface">,  SecondInterface</span><span class="oointerface">,  ThirdInterface</span> {</div>
+ </div>
+
+</div>
```
Girgias, line 76: Ditto

Girgias (approved):
Thank you this looks good :)

PR #97, opened 2024-02-13
Override hardcoded stylesheets in PHP XHTML formats

### LINKED ISSUE #70: Speed up or avoid downloading of css from http://www.php.net in render.php
When running `phd/render.php --docbook doc-base/.manual.xml --package PHP --partial=book.simdjson --format xhtml` - it takes 10 seconds

Using https://github.com/adsr/phpspy to render a flame graph:

phd-flame-graph

1. ~~Use php's curl module if available?~~
2. Use a different CDN for static assets if possible, or check if the file is recent? (This is taking 5 seconds per file, which is a different question for a different repo)
3. ~~Send etags/if-modified-since?~~ (doesn't help, still slow https://daniel.haxx.se/blog/2019/12/06/curl-speaks-etag/ )
4. It's a fast 0.01 second redirect to https://www.php.net/styles/theme-base.css 
5. Maybe check the modification time on disk and cache on disk as sha($url) on http 200s, and allow it by default if the css file modification time is less than a day old

```
Copying http://www.php.net/styles/theme-base.css
Copying http://www.php.net/styles/theme-medium.css
```

```
» wc output/php-chunked-xhtml/**/*.css  
 1884  3784 32631 output/php-chunked-xhtml/styles/03e73060321a0a848018724a6c83de7f-theme-base.css
  606  1242 11091 output/php-chunked-xhtml/styles/03e73060321a0a848018724a6c83de7f-theme-medium.css
 2490  5026 43722 total
```

```
< HTTP/2 200 
< server: myracloud
< date: Fri, 21 Oct 2022 12:52:18 GMT
< content-type: text/css
< last-modified: Mon, 17 Oct 2022 12:10:02 GMT
< etag: "634d461a-7f77"
< expires: Sat, 22 Oct 2022 11:06:08 GMT
< cache-control: max-age=86400
< x-cdn: 1
```

This might just be a wifi or cdn or isp issue, though - see https://github.com/php/web-php/issues/556

haszi:
Let the `--css` option override the default stylesheets for all PHP XHTML-based formats. Closes #70.

Phd's `--css` option allows to pass stylesheets to the rendered formats. However, in the constructor of the `PHP_ChunkedXHTML` format two stylesheets from `php.net` are added to every rendering in this format. 
This PR changes this behaviour so that the hardcoded stylesheets are only added if there are no stylesheets passed in from the command line and it applies this to all of the XHTML-based format in the PHP package. 
The default behavior (ie. without the `--css` option) does not change for the `PHP_ChunkedXHTML` format.

### DIFF
--- a/phpdotnet/phd/Package/PHP/ChunkedXHTML.php
+++ b/phpdotnet/phd/Package/PHP/ChunkedXHTML.php
@@ -8,11 +8,6 @@ class Package_PHP_ChunkedXHTML extends Package_PHP_Web {
         parent::__construct();
         $this->registerFormatName("PHP-Chunked-XHTML");
         $this->setExt(Config::ext() === null ? ".html" : Config::ext());
-
-        Config::setCss(array(
-            'http://www.php.net/styles/theme-base.css',
-            'http://www.php.net/styles/theme-medium.css',
-        ));
     }
 
     public function __destruct() {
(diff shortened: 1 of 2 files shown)


Girgias (approved):
LGTM

PR #99, opened 2024-02-16
Run tests on pull requests

haszi:
Run the few available tests on each push to the repo.

This could be used as a basis for gradually adding tests to at least the format used to generate the official docs (PHP package's XHTML format if I'm not mistaken).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,23 @@
+name: Test
+on: [push, pull_request]
+jobs:
+  test:
+    name: test
+    strategy:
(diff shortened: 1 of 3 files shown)


haszi:
And there you go, deprecation notices from 8.1 on. :-) 

Updated the failing test by adding a `xml:id` to the `<article>` element which is expected to be available by the indexer.

Girgias (commented):
I don't actually know how GitHub actions works @shivammathur could you have a sanity check look at this?

shivammathur (approved):

PR #102, opened 2024-02-22
Add links to constants

haszi:
Add links to constants with IDs in the following two standard formats (`xml:id` doesn't have to be on the `<constant>` element):

```xml
<constant xml:id="extension-namespace-or-classname.constants.name-of-constant">
 Extension\Namespace\Or\ClassName::NAME_OF_CONSTANT
</constant>

<constant xml:id="constant.name-of-constant">
 NAME_OF_CONSTANT
</constant>
```

Please note that there is a large number of constants not using one of these standard formats.


Girgias (approved):
I think this is a good initial improvement

PR #103, opened 2024-02-22
Enable nested roles

haszi:
Refactor `role` to be and array instead of a string to allow nesting of roles. Use the `indent` method to format constant list tables. Add tests for regular, table formatted and nested role containing `<variablelist>`s.

This PR fixes the shortcoming of the previous `role="constant_list"` PR and enables nested roles in `<variablelist>`s which allows to format constant lists in the documentation as tables without these being thrown off by elements that are rendered by adding roles to them (e.g. `<literal>`).

### DIFF
--- a/phpdotnet/phd/Format/Abstract/XHTML.php
+++ b/phpdotnet/phd/Format/Abstract/XHTML.php
@@ -2,7 +2,7 @@
 namespace phpdotnet\phd;
 
 abstract class Format_Abstract_XHTML extends Format {
-    public $role = false;
+    public $role = [];
 
     /* XHTMLPhDFormat */
     protected $openPara = 0;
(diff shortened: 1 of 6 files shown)


phpdotnet/phd/Format/Abstract/XHTML.php
```diff
@@ -2,7 +2,7 @@
 namespace phpdotnet\phd;
 
 abstract class Format_Abstract_XHTML extends Format {
-    public $role = false;
+    public $role = [];
```
Girgias, line 5: Add a comment that this is a FILO stack, maybe also make it private and add methods to push/pop from it where all accesses to $role must go through?

Girgias (approved):

PR #105, opened 2024-02-26
Fix GH-98

### LINKED ISSUE #98: Linking to ID on section which is not chunked does not generate a proper link
Linking to an ID that is on an element with `annotations="chunk:false"` does not work.

Related to https://github.com/php/doc-en/issues/3178

haszi:
Store IDs of non-chunked elements.
Closes #98.

### DIFF
--- a/phpdotnet/phd/Index.php
+++ b/phpdotnet/phd/Index.php
@@ -89,7 +89,7 @@ class Index extends Format
     private $currentid;
     private $chunks    = array();
     private $isChunk   = array();
-    private $nfo       = array();
+    protected $nfo       = array();
     private $isSectionChunk = array();
     private $log = '';
     private $previousId = "";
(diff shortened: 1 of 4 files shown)


Girgias (approved):
LGTM and thank you!

Girgias:
Well the tests seem to fail :thinking:

haszi:
> Well the tests seem to fail 🤔

Yeah, for some reason it doesn't index anything on GH but works perfectly fine locally. I'm trying to figure out what's going on but don't really have a clue what it could be yet.

haszi:
It turns out that `'index' => true` in `Config` doesn't do anything (changed it to `'force_index' => true`) and the `$indexes` property of `Format` is actually the list of indexes read from cache/sqlite and `Index`'s `$nfo` property holds the parsed indexes.

Anyway, the test passes now and this is ready for review.

haszi:
I've changed the private `$nfo` property to `protected` so I can access it from `TestIndex` but now I'm thinking that adding a `getNfo()` method to `Index` may be a better solution.

PR #107, opened 2024-03-01
Refactor Options Handler, Interface and Parser

haszi:
Most of this PR adds dependencies of `Options_Parser` as constructor parameters (and injects them from `render.php`) and adds parameter and return types to `Options_Handler`, `Options_Interface` and `Options_Parser`.

There are six smaller commits in this PR and it's probably easier to review it commit by commit.

### DIFF
--- a/phpdotnet/phd/Options/Handler.php
+++ b/phpdotnet/phd/Options/Handler.php
@@ -8,7 +8,7 @@ class Options_Handler implements Options_Interface
      */
     public function optionList(): array
     {
-        return array(
+        return [
             'format:'      => 'f:',        // The format to render (xhtml, pdf...)
             'noindex'      => 'I',         // Do not re-index
             'forceindex'   => 'r',         // Force re-indexing under all circumstances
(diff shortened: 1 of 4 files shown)


haszi:
Smoke test failures for PHP 7.1-7.4 are due to parameter and return types. Ie. the minimum PHP version needs a bump to 8.0+. This need is not introduced by this PR though as the codebase already makes use of PHP 8.0+ features (e.g.: `match`, `str_contains()`, etc.).

phpdotnet/phd/Options/Parser.php
```diff
@@ -91,16 +97,15 @@ private function checkOptions() {
...
+        $this->validateOptions();
 
         $args = getopt($this->getShortOptions(), $this->getLongOptions());
         if ($args === false) {
             trigger_error("Something happend with getopt(), please report a bug", E_USER_ERROR);
         }
 
         foreach ($args as $k => $v) {
-            $handler = $this->handlerForOption($k);
+            $handler = $this->handlerForOption((string) $k);
```
Girgias, line 114: Is this cast really needed?

phpdotnet/phd/Options/Handler.php
```diff
@@ -33,29 +33,31 @@ public function optionList(): array
...
     }
 
-    public function option_f($k, $v)
+    public function option_f(string $k, mixed $v): void
     {
-        if ($k == "f") {
-            return $this->option_format($k, $v);
-        } else {
-            return $this->option_outputfilename($k, $v);
+        if ($k === "f") {
```
Girgias, line 51: Might make sense to have the `"f"` be a class constant (maybe protected/private?)

phpdotnet/phd/Options/Parser.php
```diff
@@ -4,23 +4,30 @@
...
+    private Options_Interface $defaultHandler;
+    private array $packageHandlers = [];
 
-    public function __construct() {
-        $this->defaultHandler = new Options_Handler();
-        $this->packageHandlers = $this->loadPackageHandlers();
+    /**
+     * @param array<?Options_Interface> $packageHandlers
+     */
+    public function __construct(Options_Interface $defaultHandler, ?array $packageHandlers = []) {
```
Girgias, line 13: Possible idea for latter, make ` $packageHandlers` a variadic argument of type `Options_Interface` ?

phpdotnet/phd/Options/Parser.php
haszi: Hmm. You're right, strict typing is not used in any of the files and this really is not needed.

phpdotnet/phd/Options/Handler.php
haszi: I'd like to tackle these strings in a follow up PR as I think the mapping of short option, long option and handler method should be in some sort of single data structure that is more straightforward to work with (array of value objects maybe?). But this needs some more thought put into it and I'm not sure what would be the best way to go about this.

Girgias (approved):

Girgias:
Do you want to rebase the last commit into one of the prior commits for me to rebase, or just squash-merge the whole thing?

haszi:
I think a squash merge makes sense here. The only reason there are so many commits in this PR was for me to keep track of what I was doing (and possibly to make reviewing it easier). I'll squash commits like these in the future before I open a PR.

Girgias:
You can keep the multiple commits when doing PRs as it makes reviewing easier :)

PR #109, opened 2024-03-12
Add handler for caption tag

haszi:
Add handler for caption tag. The first of these tags has just been added to `doc-en` a few days ago.

### DIFF
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -26,6 +26,7 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
             'part'              => 'format_chunk',
         ),
         'book'                  => 'format_container_chunk_top',
+        'caption'               => 'format_caption',
         'chapter'               => 'format_container_chunk_top',
         'citetitle'             => 'em',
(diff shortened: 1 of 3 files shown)


phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -2280,4 +2281,7 @@ public function format_whitespace($whitespace, $elementStack, $currentDepth) {
         return $whitespace;
     }
 
+    public function format_caption($open, $name, $attrs, $props) {
+        return $open ? '<div class="caption">' : '</div>';
```
Girgias, line 2285: Isn't the `<figcaption>` tag more appropriate?` although this probably requires more change to support images being put in a figure tag?

haszi: A `<figcaption>` is a caption for a `<figure>` element but this `<caption>` is used in a `<mediaobject>` element. There is a DocBook example of a `<caption>` [here](https://tdg.docbook.org/tdg/5.2/informalfigure) the HTML of which is also using `<div>`s for `<mediaobject>` and its `<caption>`, so I just followed their lead on this. :-)

Also, I think `<figcaption>` is an HTML element and not a DocBook one.

Girgias:
Thank you!

PR #110, opened 2024-03-14
Refactor Options Parser and default Options Handler

haszi:
This PR includes the followings:
 - remove all direct setting of `Config` from the default `Options_Handler` and return an array of parsed options instead. Add return type declarations to each method returning an array
 - inject dependencies through `Options_Handler`'s constructor in `render.php`
 - refactor a few files to allow them to be used from tests
 - add tests for the default `Options_Handler` and `Options_Parser`
 - fix inconsistent short and long option required/optional flags in `Options_Handler`

### DIFF
--- a/phpdotnet/phd/Autoloader.php
+++ b/phpdotnet/phd/Autoloader.php
@@ -1,7 +1,7 @@
 <?php
 namespace phpdotnet\phd;
 
-require __DIR__ . DIRECTORY_SEPARATOR . 'Config.php';
+require_once __DIR__ . DIRECTORY_SEPARATOR . 'Config.php';
 
 class Autoloader
 {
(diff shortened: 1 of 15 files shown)


haszi:
Possible future improvements of this part of Phd:

- add a `handle()` method to `Options_Interface` (or a new `Options_Handler` interface) and implement it in the default `Options_Handler` class and simply pass CLI arguments/values to it
- refactor CLI options parsing so that these can be passed in through GET/POST parameter values as well
- refactor the default `Options_Handler`'s short option/long option/method mapping (not sure what this would look like)
- refactor the default `Options_Handler` so that it doesn't need an instance of `Config` and/or `Format_Factory` (not sure how this would work with all the settings and format info needed by the handler)

Girgias (approved):
LGTM

PR #111, opened 2024-03-15
Add links to namespaced non-class constants

haszi:
Add links to namespaced non-class constants.
Add namespaced non-class constants to tests.


Girgias (approved):
Makes sense to me

PR #112, opened 2024-03-24
Refactor Reader and Reader_Partial

haszi:
Inject dependencies through the constructor of `Reader_Partial`, add parameter and return type hints and add tests for both `Reader`s.


Girgias (approved):
Did a brief review but LGTM

PR #113, opened 2024-03-28
Refactor Index

haszi:
This PR is a minimal refactoring of `Index` to move its dependencies out of the class and inject them through the constructor.

The changes are:
 - move `requireIndexing()` method out of `Index` into a global function that receives all its dependencies through parameters
 - add a new `IndexRepository` class that handles all database access for `Index`, refactor `Index` to receive an object of this class as a constructor dependency and to pass raw data to it (i.e. data not containing SQL statements)
 - add basic indexing test
 - fix existing test that uses indexing to use `IndexRepository`

### DIFF
--- a/phpdotnet/phd/Index.php
+++ b/phpdotnet/phd/Index.php
@@ -112,50 +112,6 @@ class Index extends Format
     public function appendData($data) {
     }
 
-    /**
-     * Checks if indexing is needed.
-     *
-     * This is determined the following way:
-     * 0. If no index file exists, indexing is required.
-     * 1. If the config option --no-index is supplied, nothing is indexed
-     * 2. If the config option --force-index is supplied, indexing is required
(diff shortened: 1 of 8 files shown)


haszi:
Possible future improvements:

 - push `IndexRepository` further up the inheritance tree to `Format`, this will allow all parts of Phd to use the same indexing database (as opposed to creating new instances in multiple places and sometimes not using the same database that was used for indexing)
 - rewrite SQL in `IndexRepository` to use prepared statements and to possibly use one large `INSERT INTO` statement vs individual ones for each of the 29k+ index rows
 - remove the global `ReaderKeeper` (from all classes) and pass the current `Reader` to `Format`s as needed
 - move all index database access from `Format` to `IndexRepository`

Girgias (approved):
Did a brief review but I can't spot anything majorly wrong.

PR #118, opened 2024-04-25
Use the same indexing database throughout

haszi:
Use the same indexing database in every part of PhD by always retrieving it from `indexcache` and always set `indexcache` before any indexing or rendering is done in `render.php`.


Girgias (approved):
Cursory look, but LCTM

PR #120, opened 2024-04-26
Move all database access code to IndexRepository

haszi:
Move all database access code to `IndexRepository` and use the same `IndexRepository` instance everywhere.


Girgias (approved):
LGTM

render.php
```diff
@@ -75,30 +75,18 @@
 
 function make_reader() {
     //Partial Rendering
-    $idlist = Config::render_ids() + Config::skip_ids();
+    $idlist = $config->render_ids() + $config->skip_ids();
```
sy-records, line 78: Missing `$config`

PR #121, opened 2024-04-26
Full doc-en render test

haszi:
This could be an additional test for testing PhD with a full render of the `doc-en` repo.

### DIFF
--- a/.github/workflows/full_render_doc-en_test.yml
+++ b/.github/workflows/full_render_doc-en_test.yml
@@ -1,0 +1,29 @@
+name: Test
+on: [push, pull_request]
+jobs:
+  test:
+    name: doc-en full render test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PhD
+        uses: actions/checkout@v2
+        with:
+          path: "phd"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/full_render_doc-en_test.yml
```diff
@@ -0,0 +1,29 @@
+name: Test
+on: [push, pull_request]
+jobs:
+  test:
+    name: doc-en full render test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PhD
+        uses: actions/checkout@v2
```
Girgias, line 9: Normal this uses the V2 version?

Girgias (approved):
I think this makes sense to me, I don't really know GitHub actions, but this looks ok

.github/workflows/full_render_doc-en_test.yml
haszi: You're right, it should probably be v4. I copied it from an other test. :-)

haszi:
I also renamed the workflow to `Render test` and the test to just `doc-en`, and added the output directory to the PhD call.

Girgias (approved):

PR #123, opened 2024-04-30
Fix GH-3353 incorrect function tag linking

haszi:
Closes https://github.com/php/doc-en/issues/3353.

When the list of `refentry`s (functions and methods) is retrieved from the indexing database the characters `_`, `::` and `->` in their names were replaced by `-` before storing them in the array `$ref` in `Format.php`. As this array is used to generate the links to and the index list of functions/methods, functions and methods with the same name after replacing the before mentioned characters (e.g. `finfo_buffer()`/`finfo::buffer()`, `finfo_file()`/`finfo::file()`, etc.) were overwriting each other's reference entries in memory (in the `$ref` array).

Fix this by not replacing the above listed characters.

### DIFF
--- a/phpdotnet/phd/IndexRepository.php
+++ b/phpdotnet/phd/IndexRepository.php
@@ -189,13 +189,7 @@ SQL;
 
 
     private function SQLiteRefname($context, $index, $id, $sdesc): void {
-        $ref = strtolower(
-                str_replace(
-                    ["_", "::", "->"],
-                    ["-", "-", "-"],
-                    html_entity_decode($sdesc, ENT_QUOTES, 'UTF-8')
-                )
-            );
+        $ref = strtolower(html_entity_decode($sdesc, ENT_QUOTES, 'UTF-8'));
         $this->refs[$ref] = $id;
     }
 
--- a/tests/data/bug_doc-en_GH-3353.xml
+++ b/tests/data/bug_doc-en_GH-3353.xml
@@ -1,0 +1,41 @@
+<refentry xml:id="bug_doc-en_GH-3353">
+ <refentry xml:id="method.and.function">
+  <refnamediv>
+   <refname>ClassName::methodName</refname>
+   <refname>classname_methodname</refname>
+   <refpurpose>1. This is the first method/function description</refpurpose>
+  </refnamediv>
+  <refsect1 role="description" xml:id="method.and.function-description">
+   <para>
+    Link to <function>ClassName::methodName</function>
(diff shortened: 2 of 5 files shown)


Girgias:
Could you rebase to solve conflicts?

haszi:
If you don't mind, I'll wait for #126 to be merged before rebasing.

haszi:
Rebased and ready for review.

phpdotnet/phd/IndexRepository.php
```diff
@@ -189,13 +189,7 @@ public function getRefNames(): array {
...
 
     private function SQLiteRefname($context, $index, $id, $sdesc): void {
-        $ref = strtolower(
-                str_replace(
-                    ["_", "::", "->"],
-                    ["-", "-", "-"],
-                    html_entity_decode($sdesc, ENT_QUOTES, 'UTF-8')
-                )
-            );
+        $ref = strtolower(html_entity_decode($sdesc, ENT_QUOTES, 'UTF-8'));
```
Girgias, line 192: Should there be a "normalisation" function that we can use in the different packages?

tests/data/bug_doc-en_GH-3353.xml
```diff
@@ -0,0 +1,41 @@
...
+ </refentry>
+
+ <refentry xml:id="method.only">
+  <refnamediv>
+   <refname>ClassName::methodName</refname>
+   <refpurpose>2. This is the second (method) description</refpurpose>
+  </refnamediv>
+  <refsect1 role="description" xml:id="method.only-description">
+   <para>
+    Link to <function>ClassName::methodName</function>
```
Girgias, line 23: Can you have a test where `<methodname>` is used?

phpdotnet/phd/IndexRepository.php
haszi: I'm not sure. As you can see from my changes there are three places where these references are accessed (PHP XHTML and PDF rendering, and this where the indexes are retrieved from the database) but this instance doesn't match the other two. All three operate on the same XML document and access the same `xml:id`s on the same `<refentry>`'s. There's no issues with linking these functions/methods because (I think) this `html_entity_decode()` doesn't do anything on our valid `xml:id`s anyway.

TL;DR this line in `IndexRepository.php` should probably be `$ref = strtolower($sdesc);` like the other two lines and there's probably no normalization function needed.

Girgias (approved):

PR #126, opened 2024-05-14
Move all database access to IndexRepository

haszi:
Move all database access to IndexRepository.
Use the same database during the entire indexing-rendering process. 
Add a simple smoke test for render.php.

This PR is the same as #120 with the following exceptions:
 - `functions.php`'s `requireIndexing()` is now checking whether `Config`'s `indexcache` exists (where the indexing database used everywhere is stored in and after this PR), therefore the `SQLite` database parameter has been removed
 - `indexcache` is set for every run so setting it after indexing has been removed in `render.php`
 - added the `$config` parameter (of type `Config`) to `make_reader()` in `render.php`, calling it with `new Config` arguments at call sites (this was the issue that broke the builds in the last PR, thanks to @sy-records for pointing it out)
 - added a smoke test for `render.php` so that the last PR's issues can be caught immediately in a local dev environment too

### DIFF
--- a/phpdotnet/phd/Format.php
+++ b/phpdotnet/phd/Format.php
@@ -24,7 +24,7 @@ abstract class Format extends ObjectStorage
     private $elementmap = array();
     private $textmap = array();
     private $formatname = "UNKNOWN";
-    protected $sqlite;
+    private IndexRepository $indexRepository;
 
     protected $title;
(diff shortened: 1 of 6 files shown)


Girgias:
Hopefully this works this time :D

PR #130, opened 2024-05-22
Remove ReaderKeeper from Index

haszi:
`ReaderKeeper` is a global static object wrapper around an array that contains the active `Reader` and is used by three `Format`s to read the contents of certain tags as string, XML or return it as a DOM Node. 

This PR removes `ReaderKeeper` from the `Index` `Format` by delaying writing the long and short descriptions, and the changelog tables until after the appropriate tags are closed. This is done in a similar way to how method parameters are rendered, ie. by keeping everything inside the tag in temporary variables and processing them when the tag is closed.

There is no test for this PR but I've done a full render of `doc-en` with and without these changes and there are zero diffs for the ~12k files generated.

### DIFF
--- a/phpdotnet/phd/Index.php
+++ b/phpdotnet/phd/Index.php
@@ -83,6 +83,10 @@ class Index extends Format
         'phpdoc'            => 'PI_PHPDOCHandler',
     );
 
+    private bool $inLdesc = false;
+    private bool $inSdesc = false;
+    private string $currentLdesc = "";
+    private string $currentSdesc = "";
     private $currentchunk;
     private $ids       = array();
     private $currentid;
@@ -93,6 +97,7 @@ class Index extends Format
     private $previousId = "";
     private $inChangelog = false;
     private $currentChangelog = array();
+    private string $currentChangeLogString = "";
     private $changelog        = array();
     private $currentMembership = null;
     private $commit     = array();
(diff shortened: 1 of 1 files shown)


phpdotnet/phd/Index.php
```diff
@@ -83,6 +83,10 @@ class Index extends Format
         'phpdoc'            => 'PI_PHPDOCHandler',
     );
 
+    private bool $inLdesc = false;
+    private bool $inSdesc = false;
+    private string $currentLdesc = "";
+    private string $currentSdesc = "";
```
Girgias, line 89: For what does the L and S stand for?

haszi: Long and short. 

Long descriptions are text in `<title>` and `<refpurpose>` tags (e.g. class name descriptions (`'The DOMDocument class'`), function and method descriptions), short descriptions are text in `<refname>` and `<titleabbrev>` tags (e.g. class names (`DOMDocument`), function and method names).

Girgias: Please use those in the property name then, as I prefer a long and descriptive name than this cryptic name.

haszi: I've followed the existing naming convention but in retrospect that wasn't a smart choice. :-)

Girgias:
Thank you!

PR #132, opened 2024-05-27
Copy PhD guide into README

haszi:
Convert PhD guide from XML to markdown.
Copy markdown PhD guide into README.
Change extension of README to .md.
Remove url to non-existent page.
Add section about language repositories.


Girgias (commented):
Does it make sense to already update part of the requirements, or just merge and then do follow-up PRs?

haszi:
Since I've already made some changes, I think we could do all the updates in this PR. And I've already changed the PHP requirement from 7.1 to 8.0+. :-)

tiffany-taylor:
Nit: "an own pear channel" is awkward phrasing in English. I want to suggest "its own pear channel" but that implies a stronger connection between PhD and Pear that I'm guessing shouldn't be implied.

haszi:
> Nit: "an own pear channel" is awkward phrasing in English. I want to suggest "its own pear channel" but that implies a stronger connection between PhD and Pear that I'm guessing shouldn't be implied.

To be honest, I just copied everything from the `xml` file without reading a large part of it. :-) I've just made the change you suggested (and some more). Could you take another look?

tiffany-taylor:
Thank you for going through the rest of the phrasing in the README and improving it, it is definitely better. Reading through the phrasing of the file, there are further changes I want to make, but it will have to be tomorrow or later in the week. I'm currently tired, and I'm not confident in my ability to get all the way through the whole file, or my ability to phrase stuff clearly and succinctly. 

Aside from that, I have a question that is open-ended, probably a question for @Girgias. I'm wondering if the Installation section should be redone in such a way that we suggest cloning the repo and using the command `php phd/render.php` to render the docs before we suggest installing PhD via PEAR. I'm not certain on this, there may be good reasons for why we want to have the PEAR suggestion first.

On a different note, I can't remember if passive or active voice is encouraged within docs. I know that the docs for contributors is looser than what is expected for the manual, but if I recall correctly, if active voice can be used, it should be used?

Girgias:
Frankly I would remove mentions of PEAR as I don't think we maintain that any more.

Honestly, the "advice" to always use the active voice is kinda nonsense. The passive voice allows to move the subject of a sentence to the end for it to be the focal point.

Fun 10min video about this: https://www.youtube.com/watch?v=sS-Txm3R3v8

README.md
```diff
@@ -0,0 +1,568 @@
...
+$ pear install doc.php.net/phd_php
+Starting to download PhD_PHP-1.0.0.tgz (18,948 bytes)
+[...]
+install ok: channel://doc.php.net/PhD_PHP-1.0.0
+
+$ pear install doc.php.net/phd_pear
+downloading PhD_PEAR-1.0.0.tgz ...
+[...]
+install ok: channel://doc.php.net/PhD_PEAR-1.0.0
+```
```
Girgias, line 81: I would remove this whole block

README.md
```diff
@@ -0,0 +1,568 @@
...
+```shell
+for i in package_*.xml; do pear install $i; done
+```
+
+```shell
+$ phd --version
+PhD Version: phd-from-svn
+PHP Version: 5.3.3-dev
+Copyright(c) 2007-2010 The PHP Documentation Group
+```
```
Girgias, line 110: Actually this one as well, as I don't even know if "installation" works, and really should not be done

README.md
```diff
@@ -0,0 +1,568 @@
...
+
+### Using PhD without installation
+
+You can use PhD without installing it or PEAR by simply cloning the git repository:
```
Girgias, line 114:
```suggestion
### Using PhD 

To use PhD you must clone the git repo:
```

README.md
```diff
@@ -0,0 +1,568 @@
...
+
+Now you have everything you need. Just type
+
+```shell
+$ cd phd/docs/phd-guide/
+$ phd -f bigxhtml -d phd-guide.xml
+```
+
+There should be an .html file in the directory now. View it with a
+browser!
```
Girgias, line 331: Possibly removed, or at least the commands need to be updated because we don't use SVN any more.

README.md
```diff
@@ -0,0 +1,568 @@
...
+```shell
+$ phd -g 'phpdotnet\phd\Highlighter_GeSHi' -d phd-guide.xml
+```
+
+If you have GeSHi version 1.1.x installed, you should use the
+`phpdotnet\phd\Highlighter_GeSHi11x` highlighter, which is adapted to
+GeSHi's new API.
+
+Apart from using the highlighter shipped with PhD, you can [build your
+own highlighters](#writing-a-custom-syntax-highlighter).
```
Girgias, line 387: Does this even still work ?

README.md
```diff
@@ -0,0 +1,568 @@
...
+documentation as easy as possible, some custom tags have been added to the
+DTD.
+
+All extensions live in their own XML namespace "`phd:`" which resolves
+to <http://www.php.net/ns/phd>. When using one of the attributes or tags
+described here, remember to set the namespace:
+
+```xml
+xmlns:phd="http://www.php.net/ns/phd"
+```
```
Girgias, line 401: Not really relevant anymore.

README.md
```diff
@@ -0,0 +1,568 @@
...
+    xml:id="preface"
+    phd:chunk="false"
+>
+    <info>
+    <title>Preface</title>
+    ..
+    </info>
+    ..
+</preface>
+```
```
Girgias, line 439: We use the standard `annotation` attribute now.

README.md
```diff
@@ -0,0 +1,568 @@
...
+You can add a title with `<title>`.
+
+#### Attributes
+
+| Attribute name | Description | Default value |
+|----|----|----|
+| phd:element | ID of the element whose children shall be linked | *none* |
+| phd:toc-depth | Depth of the TOC/Number of levels | `1` |
+
+Attributes for \<phd:toc\>
```
Girgias, line 464: Does this even work? Also conflicts with standard docbook (and really should be a PI if supported)

README.md
```diff
@@ -0,0 +1,568 @@
...
+Class and variable name as `phd:linkend` text, separated by a double
+colon and a dollar sign before the variable name.
+
+```xml
+<phd:pearapi phd:package="Net_Geo" phd:linkend="Net_Geo::$cache_ttl"/>
+<phd:pearapi phd:package="Net_Geo" phd:linkend="Net_Geo::$cache_ttl">some text</phd:pearapi>
+```
+
+[» Net_Geo::$cache_ttl](https://pear.php.net/package/Net_Geo/docs/latest/apidoc/Net_Geo/Net_Geo.html#var$cache_ttl)
+[» some text](https://pear.php.net/package/Net_Geo/docs/latest/apidoc/Net_Geo/Net_Geo.html#var$cache_ttl)
```
Girgias, line 529: PEAR is outdated, so not sure if this serves any purpose any more

README.md
```diff
@@ -0,0 +1,568 @@
...
+Once you wrote your custom source code highlighting class, it's time to
+[try it out](#source-code-highlighter).
+
+# Links
+
+PhD 0.1RC1 release announcement:
+
+-   [PhD 0.1RC1
+    released](https://bjori.blogspot.com/2007/10/phd-php-based-docbook-renderer-rc1.html)
+    by Hannes Magnusson (PhD 0.1RC1)
```
Girgias, line 566: I think this can also be removed, nice piece of history but it's 17y old now

README.md
haszi: The `phd-guide.xml` file referenced here is the file that I've copied all this from so I'll just remove this section. :-)

README.md
haszi: This doesn't work without PEAR so I think we should remove this too.

README.md
haszi: This is not used in `doc-en`. I'll remove it from the README.

haszi:
I've pushed another commit with some minor changes. 

Also, I'm thinking that the section on the PhD's options isn't too helpful in its current form. It should either be expanded (ie. discussing some of the more useful options instead of showing a list of all options) or removed completely.

Girgias (approved):
I agree about the options and commands should be expanded to be more useful, but lets iterate about this in tree.

PR #137, opened 2024-06-05
Fix doc-en GH-3428

haszi:
Replace static example element counter with class properties in `Index` and `generic XHTML`. Add tests.
Refactor `TestRender` so that it can render multiple formats.

Closes [doc-en #3428](https://github.com/php/doc-en/issues/3428).


Girgias (approved):
Thank you very much for fixing this bug!

PR #127, opened 2024-05-17
Add Attribute linking

haszi:
This PR adds support for attributes in the format described in [this comment](https://github.com/php/doc-en/pull/3385#discussion_r1601401758) and generates links to the documentation pages of known predefined attributes.

*Edit:*
As a side note, there are 129 additional attributes in `php-src` stub files that are not documented in `doc-en`.

PR #140, opened 2024-06-16
Add property and constant attribute rendering

haszi:
Add class property and constant attribute rendering support.

This, the [parameter](#127) and the [class, method and function](#135) attribute PRs together close https://github.com/php/doc-en/issues/3405.


haszi:
This PR renders properties and constants in the following format:

properties_constants

kocsismate:
Thank you!

PR #141, opened 2024-06-23
Update README

haszi:
Add references to the other parts of the documentation pipeline.
Remove section about formats.
Remove some options and reformat the remaining ones.

@tiffany-taylor Pinging you, as promised. Please take a look and let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,75 +83,41 @@ on setting up a local mirror of the PHP documentation.
 
 ## PhD's rendering options
 
-Let's take a closer look at PhD and see what capabilities are available to us.
+The following sections list PhD's the most frequently used options.
+To see the list of all options run PhD with the `-h` \ `--help` option.
 
-```shell
-$ php phd/render.php --help
-PhD version: 1.1.12
-Copyright(c) 2007-2024 The PHP Documentation Group
+### Input
 
-    -v
-    --verbose <int>            Adjusts the verbosity level
-    -f <formatname>
-    --format <formatname>      The build format to use
-    -P <packagename>
-    --package <packagename>    The package to use
-    -I
-    --noindex                  Do not index before rendering but load from cache
-                                (default: false)
-    -M
-    --memoryindex              Do not save indexing into a file, store it in memory.
-                                (default: false)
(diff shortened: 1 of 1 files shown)


tiffany-taylor:
For the most part, I really like the changes that have been made. It's much easier to read through and make sense of what's going on, and the phrasing is akin to the other PHP docs tutorials. The only thing I'm not certain on, is the removal of the package and format options, but mentioning the `--list` option, as you've done, may be sufficient. Another thought is maybe having a section that lists the different packages and formats, but displayed in a table format rather than as shell output. I'll do some more thinking on this.

tiffany-taylor:
I'm still going back and forth on having these listed. The biggest con is that they can get outdated, and it's a maintenance burden. From personal experience with needing to check the complete list of options for something and I happen to be not near a shell, having a list on a web page somewhere, even if it's a little outdated has been better than none. Similar to a manpage. I will let others weigh in, I don't want to make the decision on this.

haszi:
> ...having a list on a web page somewhere, even if it's a little outdated has been better than none...

I feel the same way but of the 26 formats, I know there are at least 12 that are either throwing an uncaught exception or raising hundreds of notices and warnings. On top of that, at least the PDF formats are completely outdated. 
As far as I can tell, the only formats that are in a good shape are the PHP package's XHTML-based formats and the generic XHTML format (to a certain degree). And those are the only ones we have tests for too. 

TL;DR: other than a handful of XHTML-based formats, I don't think most of the formats are usable. :-)

Girgias (approved):

haszi:
I have added the list of supported packages and formats to the end of the README and linked to it from the appropriate parts of the output rendering options. 

@tiffany-taylor would something like this be okay?

tiffany-taylor:
I love the look of the table, but I've been trying to reach a decision on the contents of table. Should it include all of the options, despite some of them not working? That's the question I've been mulling over, and I feel like it should only have the options that currently work. :/

Girgias:
Merging as we can always decided later if we should remove the options that don't work.

PR #142, opened 2024-06-23
Add config as a dependency of format

haszi:
Add Config as a dependency to Format and all it subclasses.
Inject Config into every Format subclass at instantiation.
Fix tests.

This is a follow up to #139. As mentioned there, this PR adds `Config` as a dependency to `Format` and all its subclasses, and injects a `Config` object into every subclass of `Format` (in `render.php` and `Factory.php`).

### DIFF
--- a/phpdotnet/phd/Format/Abstract/PDF.php
+++ b/phpdotnet/phd/Format/Abstract/PDF.php
@@ -132,7 +132,8 @@ class PdfWriter {
     // To temporarily store $current(s)
     private $old = array();
 
-    function __construct($pageWidth = 210, $pageHeight = 297) {
+    function __construct($pageWidth = 210, $pageHeight = 297, Config $config) {
+        parent::__construct($config);
     	// Initialization of properties
     	$this->haruDoc = new \HaruDoc;
     	$this->haruDoc->addPageLabel(1, \HaruPage::NUM_STYLE_DECIMAL, 1, "Page ");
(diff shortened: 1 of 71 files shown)


phpdotnet/phd/Format/Abstract/PDF.php
```diff
@@ -132,7 +132,8 @@ class PdfWriter {
     // To temporarily store $current(s)
     private $old = array();
 
-    function __construct($pageWidth = 210, $pageHeight = 297) {
+    function __construct($pageWidth = 210, $pageHeight = 297, Config $config) {
```
kamil-tekiela, line 135: Deprecated: Optional parameter $a declared before required parameter $b is implicitly treated as a required parameter

phpdotnet/phd/Package/Generic/ChunkedXHTML.php
```diff
@@ -2,8 +2,8 @@
 namespace phpdotnet\phd;
 
 class Package_Generic_ChunkedXHTML extends Package_Generic_XHTML {
-    public function __construct() {
-        parent::__construct();
+    public function __construct(Config $config) {
```
kamil-tekiela, line 5: You forgot to pass the parameter during initialization in `Package_IDE_Base`

phpdotnet/phd/Format/Abstract/PDF.php
haszi: You're right. What do you think about moving `$pageWidth` and `$pageHeight` to `Config`? These are configuration details so should probably be properties of `Config`, right?

kamil-tekiela: Yes or you could make the Cofig as the first parameter

haszi: I've just realized that `PdfWriter` is not even a `Format` so doesn't need an instance of `Config` at all. I'm removing it from the parameter list and removing the next line as well.

PR #146, opened 2024-08-19
Fix element map so changelog rows are handled correctly

jimwins:
https://www.php.net/manual/en/doc.changelog.php is empty because this was backwards and changelog entries weren't actually being indexed.

### DIFF
--- a/phpdotnet/phd/Index.php
+++ b/phpdotnet/phd/Index.php
@@ -98,7 +98,7 @@ class Index extends Format
     private $inChangelog = false;
     private $currentChangelog = array();
     private string $currentChangeLogString = "";
-    private $changelog        = array();
+    protected $changelog       = array();
     private $currentMembership = null;
     private $commit     = array();
     private $POST_REPLACEMENT_INDEXES = array();
--- a/phpdotnet/phd/Index.php
+++ b/phpdotnet/phd/Index.php
@@ -419,17 +419,19 @@ class Index extends Format
             $this->currentChangelog[] = htmlentities(trim($this->currentChangeLogString), ENT_COMPAT, "UTF-8");
         }
     }
+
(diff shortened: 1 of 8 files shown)


cmb69:
Hmm, wouldn't that revert bacc87c11a1b209b1a51c201824533558d66ac22? @sy-records can you check?

jimwins:
Yeah, I'm really not quite sure what is going here. Without my change, `format_row()` never gets called so nothing is added to the `$this->changelog` array. But with the change, some things in rows don't get properly linked.

Looks like we have one map that is trying to handle `<tbody><row><entry>` and another just `<tbody><row>` and things are getting confused. `format_entry()` is treating `$this->currentChangelog` as an array of strings, `format_row()` treats it as an array containing of the membership and id?

sy-records (approved):
I tested it. 

LGTM.

jimwins:
Pushed a change to the PR to do that version splitting. Together with the cleanup in php/doc-en#3672, this results in a `doc.changelog.html` that is 329K, but that includes changes since 5.0.0. Changing it to 8.0.0 takes it down to 274K. (This is configured in `doc-base/manual.xml/in`.)

cmb69:
> […], but that includes changes since 5.0.0.

Actually, the current documentation shouldn't care about PHP 5 anymore.

jimwins:
Opened php/doc-base#146 to bump up it up to 7.0.0. That shaves off ~8K.

haszi:
Could you also add a test showing how the changelog entries are now indexed and another one that shows that whatever was not properly linked after the first commit is now being linked correctly?

haszi (approved):

jimwins:
Will merge this once php/doc-en#3672 is approved and merged, I'd rather not merge this and end up with the messy changelog that results without those cleanups.

PR #148, opened 2024-08-31
Ignore leading/trailing space in function/method/variable names when linking (closes #80)

### LINKED ISSUE #80: Be more lax in function/methodnames tag in regards to whitespace
I recently fixed an issue on pt_BR where linking on `<function>` was failing because whitespace inside the tag. So I suggest trim() and strtolower() the tag's content before calculating link targets.

_Originally posted by @alfsb in https://github.com/php/doc-en/issues/2658#issuecomment-1669890373_

jimwins:
This doesn't quite work, I think the XML in my test case isn't right because the class name isn't actually get linked and I can see that it's not getting indexed correctly.

### DIFF
--- a/phpdotnet/phd/Format.php
+++ b/phpdotnet/phd/Format.php
@@ -270,12 +270,15 @@ abstract class Format extends ObjectStorage
         return $this->examples;
     }
     public function getRefnameLink($ref) {
+        $ref = trim($ref);
         return isset($this->refs[$ref]) ? $this->refs[$ref] : null;
     }
     public function getClassnameLink($class) {
+        $class = trim($class);
         return isset($this->classes[$class]) ? $this->classes[$class] : null;
     }
     public function getVarnameLink($var) {
+        $var = trim($var);
         return isset($this->vars[$var]) ? $this->vars[$var] : null;
     }
     public function getGeneratedExampleID($index) {
--- a/tests/render/bug_GH-80.phpt
+++ b/tests/render/bug_GH-80.phpt
@@ -1,0 +1,103 @@
+--TEST--
+GH-80: Be more lax in function/methodnames tag in regards to whitespace
+--FILE--
+<?php
+namespace phpdotnet\phd;
+
+require_once __DIR__ . "/../setup.php";
+
+$xml_file = __DIR__ . "/data/bug_GH-80.xml";
+
+$config->setXml_file($xml_file);
+$config->setForce_index(true);
+
+$render = new Render();
+
+$indexRepository = new IndexRepository(new \SQLite3(":memory:"));
+$indexRepository->init();
+$config->set_indexcache($indexRepository);
+
+$index = new TestIndex($indexRepository, $config);
+$render->attach($index);
+
+$reader = new Reader;
+$reader->open($config->xml_file(), null, LIBXML_PARSEHUGE | LIBXML_XINCLUDE);
+$render->execute($reader);
+
+$render->detach($index);
+
+$format = new TestPHPChunkedXHTML($config);
+$render = new TestRender(new Reader, $config, $format);
+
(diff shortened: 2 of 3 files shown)


haszi:
@jimwins The top level xml element needs a DocBook xml namespace declaration (ie. `xmlns="http://docbook.org/ns/docbook"`) for PhD to pick up the `role` attribute correctly.

haszi:
@Girgias We've talked about this when I wanted to fix this issue and IIRC you preferred fixing the markup in the doc repos instead of silently trimming them.

alfsb:
If it is preferable to fix the sources instead silently trimming, then this code can be changed to verbosely generate alerts where some `$target != trim( $target );` on rendering process, to make this type of linking fail easier to catch.

And/or re-implement these types of tests on doc-base scripts, as `configure.php` or QA tools does not generate alerts of this.

alfsb:
Or to not trim silently, but trim *and* generate warnings.

jimwins:
Adding the namespace declaration doesn't fix the test, it's still not indexing the name of the class for some reason. How that is supposed to get pulled from the contents is still dark magic to me.

Making this class of problem in the XML fail when building instead of just handling it is probably the better solution. Warnings will eventually just get ignored and start piling up.

alfsb:
> Making this class of problem in the XML fail when building instead of just handling it is probably the better solution. Warnings will eventually just get ignored and start piling up.

I think that adding warnings in PHD is still good, as linking fails may occurs for other reasons besides whitespace.

About warning/failing this in XML build/configure, I have a old code from when `doc-base/scripts/translation` was still a remote dream, that outputs this:

```
qaxml.w: en/reference/outcontrol/user-level-output-buffers.xml

  methodname handler

qaxml.w: en/reference/reflection/reflectionreference/fromarrayelement.xml

  classname ReflectionReference
```

haszi:
> Adding the namespace declaration doesn't fix the test, it's still not indexing the name of the class for some reason. How that is supposed to get pulled from the contents is still dark magic to me.

Try adding `<titleabbrev>Example</titleabbrev>` before the `<classsynopsis>` element. This might cause an issue with whitespace but should fix the indexing issue.

jimwins:
Okay, found the problem with my test, and it's because the way the anchors are captured for `<classname>` in `phpdotnet/phd/Package/PHP/XHTML.php` based on the rules at line 151. (It uses the `<titleabbrev>`, not anything to do with the `<classsynopsis>` like I had expected.)

We could print warnings about missing links in the `get*Link()` methods in `phpdotnet/phd/Format.php`. That prints 6,687 warnings for `doc-en` right now. I haven't tried to do any sort of breakdown, but I think a substantial fraction of them are probably expected (like references to old functions in migration chapters, or un-namespaced classes within the documentation for extensions).

haszi:
> > Making this class of problem in the XML fail when building instead of just handling it is probably the better solution. Warnings will eventually just get ignored and start piling up.
> 
> I think that adding warnings in PHD is still good, as linking fails may occurs for other reasons besides whitespace.

I agree with warning/failing in `doc-base/configure.php` but disagree with doing this in PhD as there are a lot of non-linking deprecated and removed constants and functions in the migration guides which would generate hundreds or more warnings on every run of PhD.

Girgias:
> @Girgias We've talked about this when I wanted to fix this issue and IIRC you preferred fixing the markup in the doc repos instead of silently trimming them.

I didn't have much preference but @salathe IIRC preferred fixing the XML sources.

I'm okay with allowing the linking, but it should warn so that the sources can be fixed.

alfsb:
Rephrasing what I said above, when `get*Link()` detects that a trimmed argument succeeds while a raw argument fails, a warning should be issued, so the sources should be fixed.

PR #149, opened 2024-09-01
Move removeDir() function into Web.php

haszi:
`Web.php` is the only place this function is every used in so move it there as a private method.


jimwins:
LGTM.

haszi:
Given the simple nature of these changes and if there are no objections, I'd like to merge this by the end of next week.

Girgias (approved):

PR #150, opened 2024-09-01
Show actual default PhD settings

haszi:
Show the actual default PhD settings instead of hardcoding these.


haszi:
If there are no objections by then, I'd like to merge this by the end of next week.

Girgias (approved):

PR #151, opened 2024-09-01
Move requireIndexing() to Config

haszi:
This may not make sense semantically as `Config` doesn't ever require indexing (the DocBook that is one its properties does or doesn't) so I would like to see what others think. Leaving it as a standalone function or making it an instance method of Config makes more sense?


jimwins:
I think making it an instance method of Config makes sense, maybe renaming it to `requiresIndexing()`. That turns `$config->requiresIndexing()` into a question that reads pretty naturally to me.

Girgias (approved):
MSTM

haszi:
If there are no objections, I'd like to merge this PR by the end of next week.

PR #154, opened 2024-10-04
Improve search index generation for PHP.net

lhsazevedo:
> [!NOTE]
> This is a companion PR to https://github.com/php/web-php/pull/1084, but it is not dependent on it and can be merged independently.

## Intro
PHD index entries can have both short (sdesc) and long (ldesc) descriptions. Often, the short description is empty, leading to the current fallback mechanism in `getShortDescription` and `getLongDescription`:

```php
// phpdotnet/phd/Format.php
final public function getLongDescription($id, &$isLDesc = null) {
    if ($this->indexes[$id]["ldesc"]) {
        $isLDesc = true;
        return $this->indexes[$id]["ldesc"];
    } else {
        $isLDesc = false;
        return $this->indexes[$id]["sdesc"];
    }
}
final public function getShortDescription($id, &$isSDesc = null) {
    if ($this->indexes[$id]["sdesc"]) {
        $isSDesc = true;
        return $this->indexes[$id]["sdesc"];
    } else {
        $isSDesc = false;
        return $this->indexes[$id]["ldesc"];
    }
}
```

## Problem
The current search index JSON generation doesn't utilize this fallback
mechanism, resulting in missing entries in PHP.net search results.

## Solution
This PR addresses the issue by:

- Using the long description as the title when the short description is empty.
- Using the parent book title as the long description in such cases.
- Ignoring non-page entries (non-chunk entries) when generating the search index JSON.

## Impact

- Improves search result completeness and accuracy.
- Reduces search index size by excluding irrelevant entries.

## Examples

Currently, "PHP Manual > Language Reference > Types > String" is missing from search results due to an empty short description. This change will use "Strings" (the long description) as the title and "Language Reference" (the parent book title) as the long description.

search-index.json
```diff
 [
-  "",
+  "Strings",
   "language.types.string",
   "sect1"
 ],
-[
-  "",
-  "language.types.string",
-  "sect2"
-],

...
  [
-   "",
+   "Enumerations",
    "language.types.enumerations",
    "sect1"
  ],
```

search-description.json
```diff
-  "language.types.string": "Strings",
+  "language.types.string": "Language Reference",
...
-  "language.types.enumerations": "Enumerations",
+  "language.types.enumerations": "Language Reference",
```

# Statistics
| stat | before | after | change
--- | --- | --- | ---
Entries count | 29,765 | 11,256 | -62%
Entries lacking title (sdesc) | 19,955 | 0 | -100%
Search index size | 1,383KB | 693KB | -50%

## Generated Search Index Diff Preview
Only the first 100 lines are shown for brevity.

```diff
 [
   [
-    "",
+    "Copyright",
     "copyright",
     "legalnotice"
   ],
   [
-    "",
-    "index",
-    "info"
-  ],
-  [
-    "",
-    "index",
-    "book"
-  ],
-  [
-    "",
-    "preface",
-    "section"
-  ],
-  [
-    "",
+    "Preface",
     "preface",
     "preface"
   ],
   [
-    "",
-    "intro-whatis",
-    "example"
-  ],
-  [
-    "",
+    "What is PHP?",
     "intro-whatis",
     "section"
   ],
   [
-    "",
+    "What can PHP do?",
     "intro-whatcando",
     "section"
   ],
   [
-    "",
+    "Introduction",
     "introduction",
     "chapter"
   ],
   [
-    "",
+    "What do I need?",
     "tutorial.requirements",
     "section"
   ],
   [
-    "",
-    "tutorial.firstpage",
-    "example"
-  ],
-  [
-    "",
-    "tutorial.firstpage",
-    "example"
-  ],
-  [
-    "",
+    "Your first PHP-enabled page",
     "tutorial.firstpage",
     "section"
   ],
   [
-    "",
-    "tutorial.useful",
-    "example"
-  ],
-  [
-    "",
-    "tutorial.useful",
-    "example"
-  ],
-  [
-    "",
-    "tutorial.useful",
-    "example"
-  ],
-  [
-    "",
+    "Something Useful",
     "tutorial.useful",
     "section"
   ],
   [
-    "",
-    "tutorial.forms",
-    "example"
-  ],
```

### DIFF
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -224,6 +224,28 @@ contributors($setup);
         $ids = array();
         $desc = array();
         foreach($this->indexes as $id => $index) {
+            // Skip indexes that are not chunks
+            if (!$index["chunk"]) {
+                continue;
+            }
+
+            // Handle items with long description but no short description
+            if (empty($index["sdesc"]) && !empty($index["ldesc"])) {
+                // Use long desc as short desc
+                $index["sdesc"] = $index["ldesc"];
+
+                // Find the parent book and use its long desc
+                $parentId = $index['parent_id'];
+                while (isset($this->indexes[$parentId])) {
+                    $parent = $this->indexes[$parentId];
+                    if ($parent['element'] == 'book') {
+                        $index["ldesc"] = Format::getLongDescription($parent['docbook_id']);
+                        break;
+                    }
+                    $parentId = $parent['parent_id'];
+                }
+            }
+
             $ids[] = array($index["sdesc"], $index["filename"], $index["element"]);
             $desc[$id] = $index["ldesc"];
         }
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -236,7 +236,7 @@ contributors($setup);
                 // elements (no parent) or in case the index structure is broken.
                 while (isset($this->indexes[$parentId])) {
                     $parent = $this->indexes[$parentId];
-                    if ($parent['element'] == 'book') {
+                    if ($parent['element'] === 'book') {
                         $index["ldesc"] = Format::getLongDescription($parent['docbook_id']);
                         break;
                     }
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -228,7 +228,7 @@ contributors($setup);
                 continue;
             }
 
-            if (empty($index["sdesc"]) && !empty($index["ldesc"])) {
+            if ($index["sdesc"] === "" && $index["ldesc"] !== "") {
                 $index["sdesc"] = $index["ldesc"];
 
                 $parentId = $index['parent_id'];
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -232,6 +232,8 @@ contributors($setup);
                 $index["sdesc"] = $index["ldesc"];
 
                 $parentId = $index['parent_id'];
+                // isset() to guard against undefined array keys, either for root
+                // elements (no parent) or in case the index structure is broken.
                 while (isset($this->indexes[$parentId])) {
                     $parent = $this->indexes[$parentId];
                     if ($parent['element'] == 'book') {
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -224,17 +224,13 @@ contributors($setup);
         $ids = array();
         $desc = array();
         foreach($this->indexes as $id => $index) {
-            // Skip indexes that are not chunks
             if (!$index["chunk"]) {
                 continue;
             }
 
-            // Handle items with long description but no short description
             if (empty($index["sdesc"]) && !empty($index["ldesc"])) {
-                // Use long desc as short desc
                 $index["sdesc"] = $index["ldesc"];
 
-                // Find the parent book and use its long desc
                 $parentId = $index['parent_id'];
                 while (isset($this->indexes[$parentId])) {
                     $parent = $this->indexes[$parentId];


phpdotnet/phd/Package/PHP/Web.php
```diff
@@ -224,6 +224,28 @@ protected function writeJsonIndex() {
         $ids = array();
         $desc = array();
         foreach($this->indexes as $id => $index) {
+            // Skip indexes that are not chunks
+            if (!$index["chunk"]) {
+                continue;
+            }
+
+            // Handle items with long description but no short description
+            if (empty($index["sdesc"]) && !empty($index["ldesc"])) {
```
kamil-tekiela, line 233: What is the purpose of `empty` here? Is it really possible that the key is undefined? The existing code below seems to suggest that this cannot be the case.

phpdotnet/phd/Package/PHP/Web.php
```diff
@@ -224,6 +224,28 @@ protected function writeJsonIndex() {
...
+                // Use long desc as short desc
+                $index["sdesc"] = $index["ldesc"];
+
+                // Find the parent book and use its long desc
+                $parentId = $index['parent_id'];
+                while (isset($this->indexes[$parentId])) {
```
kamil-tekiela, line 239: A code comment would make sense here to explain why `isset` is needed.

phpdotnet/phd/Package/PHP/Web.php
```diff
@@ -224,6 +224,28 @@ protected function writeJsonIndex() {
...
+                    $parent = $this->indexes[$parentId];
+                    if ($parent['element'] == 'book') {
```
kamil-tekiela, line 241:
```suggestion
                    if ($parent['element'] === 'book') {
```

phpdotnet/phd/Package/PHP/Web.php
lhsazevedo: Thanks for the quick review.

Looks like both `sdesc` and `ldesc` are initialized as empty strings:
https://github.com/php/phd/blob/417b06655c89b5bde5794dee58ce4e41834c6ef0/phpdotnet/phd/Index.php#L209-L213

However, similar `empty()` checks exist elsewhere:
https://github.com/php/phd/blob/417b06655c89b5bde5794dee58ce4e41834c6ef0/phpdotnet/phd/Index.php#L360
https://github.com/php/phd/blob/417b06655c89b5bde5794dee58ce4e41834c6ef0/phpdotnet/phd/Index.php#L372
https://github.com/php/phd/blob/417b06655c89b5bde5794dee58ce4e41834c6ef0/phpdotnet/phd/Package/PEAR/Web.php#L73-L77

So I decided to just play it safe and use `empty` as well.

For consistency, I suggest opening a separate issue to review these checks project-wide (it's hacktoberfest!). I can also address just this instance if preferred, it shouldn't break anything. What are your thoughts on the best approach?

phpdotnet/phd/Package/PHP/Web.php
```diff
@@ -224,6 +224,28 @@ protected function writeJsonIndex() {
...
+            if (!$index["chunk"]) {
+                continue;
+            }
+
+            // Handle items with long description but no short description
+            if (empty($index["sdesc"]) && !empty($index["ldesc"])) {
+                // Use long desc as short desc
+                $index["sdesc"] = $index["ldesc"];
+
+                // Find the parent book and use its long desc
```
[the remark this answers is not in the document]
lhsazevedo: Okay, removed as they weren't strictly necessary.

phpdotnet/phd/Package/PHP/Web.php
lhsazevedo: Done

phpdotnet/phd/Package/PHP/Web.php
kamil-tekiela: For consistency with the current function, let's not use `empty` here. Yes, the rest can be refactored at a later point of time.

lhsazevedo: Done, and issue created #158

kamil-tekiela:
Will this break the search if we merge it now?

lhsazevedo:
No. It is safe to merge as it doesn't alter the JSON structure.

kamil-tekiela (approved):

lhsazevedo:
Thank you!

kamil-tekiela:
Just curious. You said it would not affect the current functionality, right? Does that mean you changed something in the other PR that will avail of this change?

lhsazevedo:
I said that it wouldn't break the current search, but you can already see the improved results on php.net.

Unfortunately, the current UI hides them under the last result group ("Other Matches"). You'll probably need to scroll down the menu to see it. You may also need to clear you local storage because the search index is cached for two weeks. 

Try searching for "syntax", "types" or "operators". You should see some language reference results under "Other Matches". Those pages were missing from the index before because they don't have an `sdesc`.

haszi:
> ## Problem
> The current search index JSON generation doesn't utilize this fallback mechanism, resulting in missing entries in PHP.net search results.
> 
> ## Solution
> This PR addresses the issue by:
> 
> ...
> * Ignoring non-page entries (non-chunk entries) when generating the search index JSON.

One question:  how does removing non-chunk entries address the issue of empty long/short descriptions not using their short/long counterpart? Beside that this change not related to the problem, it also removes the possibility of searching for any useful sections that don't necessarily need an entire page (e.g. the ternary or the null coalescing operators).

lhsazevedo:
Thanks for bringing this up.

The main reason for skipping non-chunk entries was to avoid adding an `sdesc` to them, which would make them appear in the search results. The front-end does not properly handle non-chunk entries, and displaying them would lead to broken links. I realize I should have clarified this in the PR.

To properly integrate non-chunk entries into the search, we'd need to:

1. Include the container chunk ID in the index (unless `parent_id` is guaranteed to be a chunk).
2. Update the front-end to handle non-chunk links correctly, using the container chunk ID as the path and the entry ID as a hash property. For example, https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary .
3. Determine the best value for the `ldesc` for non-chunk entries. Using the container chunk's `sdesc` might be preferable over the parent book/set `sdesc`.

Your question led me to some interesting discoveries:

1. There are 18,205 non-chunk entries currently missing `sdesc`, but 370 non-chunk entries actually have both `sdesc` and `ldesc`. These are `<refentry>` aliases (like `DateTime::getTimezone`, an alias for `date_timezone_get()`) and were removed from the index after the PR. I think they should be using the same `chunk` value as the entry they alias to, but it is hardcoded as `false` (`0`):
   https://github.com/php/phd/blob/ec491540aa975c09385d0ab55c46b61361ae0c18/phpdotnet/phd/Index.php#L254-L269
   I'll create an issue so we can discuss how to handle this.

2. Another issue: while the entries' IDs are not unique, I've seen some search related code assuming they are, which definitely needs to be corrected. Will create an issue for this as well.

lhsazevedo:
@kamil-tekiela would you mind adding the [`hacktoberfest-accepted`](https://hacktoberfest.com/participation/) label to this one?

Girgias:
> @kamil-tekiela would you mind adding the [`hacktoberfest-accepted`](https://hacktoberfest.com/participation/) label to this one?

Done

PR #162, opened 2024-10-13
Bump PHP version to 8.0.0

kamil-tekiela:
The code already uses `match` construct so the minimum version must be PHP 8.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         }
     ],
     "require": {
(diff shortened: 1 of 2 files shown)


cmb69 (approved):
Indeed! See https://github.com/php/phd/blob/53f9afe269e74228017a3ee7285ffd2cf719f9cc/README.md?plain=1#L17.

PR #165, opened 2024-10-20
Fix constructor parameter issues

haszi:
Pass a required parameter and remove an unnecessary parameter and method call.


kamil-tekiela (approved):
Thank you!

PR #169, opened 2024-10-22
Bump minimum PHP version to 8.1

kamil-tekiela:
The code uses the `never` type which was introduced in PHP 8.1 so we need to bump the minimum version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ and what the different parts of its pipeline are, please refer to the
 
 ## Requirements
 
-- PHP 8.0+
+- PHP 8.1+
 - DOM, libXML2, XMLReader and SQLite3.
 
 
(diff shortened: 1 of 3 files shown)


jimwins (approved):

cmb69 (commented):
Bumping the version is okay for me, but an alternative could be to remove the `never` return type (could be an annotation instead).

And we should probably use some tooling which can detect such issues.

kamil-tekiela:
> Bumping the version is okay for me, but an alternative could be to remove the `never` return type (could be an annotation instead).
> 
> And we should probably use some tooling which can detect such issues.

PHP 8.1 is pretty old by now so I feel dropping the support is better. If I find something from PHP 8.2, I will do that.

derickr:
FYI: All the PHP servers should run PHP 8.2, so don't go using 8.3 features yet :-)

PR #175, opened 2024-11-03
Add links to constants with replaceable tags

haszi:
Add links to constants with replaceable tags by processing them when the constant tag is being closed. Also add a test.

This currently only works with one set of `<replaceable>` tags per constant (due to issues making the regex non-greedy). In practice, this shouldn't be an issue as we have no constants with more than one set of these tags.

The implementation feels a little hacky but more or less matches the processing of all other tags where post-processing is needed (all the class-based `synposis*` tags and their children).


Girgias (approved):
Thank you!

PR #179, opened 2024-11-17
Support enumsynopsis in the most minimal way

### LINKED ISSUE #89: enum rendering.
Random\IntervalBoundary is probably the first pre-defined enum, maybe we need to get PhD to render the enum?

https://github.com/php/php-src/blob/PHP-8.3/ext/random/random.stub.php#L158-L163

Girgias:
@haszi to try and resolve #89 and https://github.com/php/doc-en/pull/3419


Girgias:
This also does not support linking for `<enumname>` or `<type>` tags.

PR #182, opened 2024-12-22
Refactor Config

haszi:
This PR refactors the `Config` class from using an array of options and method calls to access these to mostly using object properties. Probably best reviewed commit by commit.

### DIFF
--- a/phpdotnet/phd/Config.php
+++ b/phpdotnet/phd/Config.php
@@ -7,61 +7,61 @@ class Config
     public readonly string $copyright;
 
     /** @var array<string, string> */
-    public array $output_format = [];
-    public bool $no_index = false;
-    public bool $force_index = false;
-    public bool $no_toc = false;
-    public string $xml_root = '.';
-    public string $xml_file = './.manual.xml';
-    public string $history_file = './fileModHistory.php';
-    public string $lang_dir = './';
+    public array $outputFormat = [];
+    public bool $noIndex = false;
+    public bool $forceIndex = false;
+    public bool $noToc = false;
+    public string $xmlRoot = '.';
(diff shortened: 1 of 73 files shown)


phpdotnet/phd/Config.php
```diff
@@ -5,49 +5,51 @@ class Config
...
-        'package_dirs'      => array(__INSTALLDIR__),
-        'saveconfig'        => false,
-        'quit'              => false,
-        'indexcache'        => '',
-        'memoryindex'       => '',
-    );
-
-    private static $optionArray = null;
+    /** @var array<string, string> */
+    public array $output_format = [];
```
Girgias, line 10: Is the change from static to instance properties okay? (not that I think having it static in the first place made a lot of sense).

Girgias (approved):
MSTM, I think it would be better to use camelCase for the property names, but this can wait.

phpdotnet/phd/Config.php
haszi: I think it makes sense that one set of configuration options can be assigned to one object of the class Config and then we just pass the appropriate object when needed.

As far as how Config is used in PhD, this doesn't make much of a difference as only one Config object is used throughout. :-)

haszi:
> MSTM, I think it would be better to use camelCase for the property names, but this can wait.

I wanted to do this eventually but it makes more sense to do this as part of this PR.

Girgias (approved):
Let me know if you are happy as if for me to merge it as a rebase or if you want to stilll organise the commits.

haszi:
> Let me know if you are happy as if for me to merge it as a rebase or if you want to stilll organise the commits.

I'll leave it up to you to merge it as you want.

Issue #164, opened 2024-10-18
Support linking from <property>


jimwins:
We do have a property tag which would be more suited here
```suggestion
   Alternatively, the <property>OAuth::$sslChecks</property> property can be set to a
```

Sadly this does not support linking yet, maybe a new project for PhD/Web for next year that @haszi (or someone else) could tackle.

_Originally posted by @Girgias in https://github.com/php/doc-en/pull/3243#discussion_r1805499390_

cmb69:
While automatic linking (like for `<function>`) would be desireable, instead of doing changes like https://github.com/php/doc-en/pull/3884, I'd rather add support for the `linkend` attribute of `property` elements.

<details><summary>POC</summary>

````diff
 phpdotnet/phd/Package/PHP/XHTML.php | 18 ++++++++++++++++++
 1 file changed, 18 insertions(+)

diff --git a/phpdotnet/phd/Package/PHP/XHTML.php b/phpdotnet/phd/Package/PHP/XHTML.php
index 16fc41b..eb39bf9 100644
--- a/phpdotnet/phd/Package/PHP/XHTML.php
+++ b/phpdotnet/phd/Package/PHP/XHTML.php
@@ -24,6 +24,7 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
         'phpdoc:classref'       => 'format_class_chunk',
         'phpdoc:exceptionref'   => 'format_class_chunk',
         'phpdoc:varentry'       => 'format_varentry_chunk',
+        'property'              => 'format_property_cmb',
         'refentry'              => 'format_refentry',
         'reference'             => 'format_reference',
         'refpurpose'            => 'format_refpurpose',
@@ -362,6 +363,23 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
         return $retval;
     }
 
+    public function format_property_cmb($open, $tag, $attrs, $props) {
+        $retval = '';
+        if ($open) {
+            if (isset($attrs[Reader::XMLNS_DOCBOOK]["linkend"])) {
+                $link = $this->createLink($attrs[Reader::XMLNS_DOCBOOK]["linkend"]);
+                $retval .= '<a href="' . $link . '">';
+            }
+            $retval .= '<code>';
+        } else {
+            $retval .= '</code>';
+            if (isset($attrs[Reader::XMLNS_DOCBOOK]["linkend"])) {
+                $retval .= '</a>';
+            }
+        }
+        return $retval;
+    }
+
[...]

PR #183, opened 2024-12-23
Add property linking

haszi:
Add linking to `<property>` elements. 

This works the same way `<constant>` tag linking works and is based on the understanding that `<property>` elements are only to be used for class properties. As with constant linking, property linking  requires the `<property>` elements to include a FQN (eg. `<property>DOMNameSpaceNode::$parentElement</property>` instead of just `<property>$parentElement</property>`) and an element with the appropriate xml ID (eg. `<varlistentry xml:id="domnamespacenode.props.parentelement">`).

Reference issue: https://github.com/php/phd/issues/164

### DIFF
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -462,6 +462,9 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
             /* DEFAULT */          false,
             'footnote'             => 'format_footnote_para_text',
         ),
+        'property'              => [
+           /* DEFAULT */           'format_property_text',
+        ],
         /* FIXME: This is one crazy stupid workaround for footnotes */
         'constant'              => array(
             /* DEFAULT */          'format_constant_text',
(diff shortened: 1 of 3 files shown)


phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -1843,6 +1846,30 @@ public function format_replaceable($open, $name, $attrs, $props) {
...
     }
+    
+    public function format_property_text($value, $tag) {
+        if (! str_contains($value, '::')) {
+            return $value;
+        }
+        
+        $tempLinkValue = str_replace(
+            array("\\", "_", "$"),
+            array("-", "-", ""),
```
Girgias, line 1857:
```suggestion
            ["\\", "_", "$"],
            ["-", "-", ""],
```

tests/package/php/data/property_linking.xml
```diff
@@ -0,0 +1,25 @@
...
+  <para>2. Nonexistent properties</para>
+  <para>
+   <property>Vendor\Namespace::$this_does_not_exist</property>
+  </para>
+ </section>
+
+ <section>
+  <para>3. Properties with leading and trailing underscores in ID</para>
+  <para>
+   <property>Extension\Class::$__leading_and_trailing_undescores__</property>
```
Girgias, line 21: Can you add examples for `camelCase` properties?

Girgias (commented):
Thank you! Some minor questions and nits, but how does it handle `<replaceable>` tags? (Because I think how PhD renders function tags which have `<replaceable>` is quite broken).

haszi:
> Thank you! Some minor questions and nits, but how does it handle `<replaceable>` tags? (Because I think how PhD renders function tags which have `<replaceable>` is quite broken).

`<replaceable>` tags won't work in properties. I thought about making it work but unlike constants, I could not think of a good reason to reference a group of class properties in this way.

On why nested tags are an issue:
PhD is a stream based renderer so every tag and their content are assumed to be self-contained by default. The bolted on functionality we use for functions, method, class- and enum synopses, etc. break that assumption and we have to do some additional workarounds to make nested tags work (as is the case with `<constant>`s with nested `<replacabeable>` tags). This is the same additional workaround that needs to be applied to `<function>` and `<methodname>` tags with `<replaceable>` parts so those render properly as well.

Girgias (approved):
Right, the stream based notion is somewhat annoying.

Issue #170, opened 2024-10-23
Linking for FQN of Classes/Methods does not work


Girgias:
Visible on the 8.4 migration guide that is live

`<classname>\Random\Randomizer</classname>` and `<methodname>\Random\Randomizer::getBytes</methodname>` do not generate links.

cmb69:
Maybe drop the leading backslashes (for now).

haszi:
From a cursory look, no linking works with FQNs. This should probably be easy to fix by trimming the leading backslash though. Or so I hope. :-)

PR #185, opened 2025-01-01
Fix FQN class, method, property and constant linking

haszi:
Minimal fix for FQN class and method/function linking not working. Closes GH-170.

### DIFF
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -1811,7 +1811,7 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
         $tempLinkValue = str_replace(
             array("\\", "_"),
             array("-", "-"),
-            strtolower(trim($constantName, "_"))
+            trim($this->normalizeFQN($constantName), "_")
         );
         
         if (str_contains($constantName, '::')) {
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -1855,7 +1855,7 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
         $tempLinkValue = str_replace(
             ["\\", "_", "$"],
             ["-", "-", ""],
-            strtolower(trim($value, "_"))
+            trim($this->normalizeFQN($value), "_")
         );
         
         list($extensionAndClass, $property) = explode("::", $tempLinkValue);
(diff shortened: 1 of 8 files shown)


phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -666,7 +666,7 @@ public function format_type_methodparam_text($type, $tagname) {
     }
 
     public function format_type_text($type, $tagname) {
-        $t = strtr(strtolower($type), ["_" => "-", "\\" => "-"]);
+        $t = strtr(ltrim(strtolower($type), "\\"), ["_" => "-", "\\" => "-"]);
```
Girgias, line 669: Maybe we should have a function/method which does the ltrim + strtolower ?

alfsb: A "normalize()" function that trims/lowers/replaces.

haszi:
Added a method for normalizing the FQNs. As it was trivial, the same fix was applied to constants and properties.

Girgias (approved):

phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -1811,7 +1811,7 @@ private function convertConstantNameToId(string $constantName): string {
         $tempLinkValue = str_replace(
```
alfsb, line 1811: In a future version, move all transformations inside the normalize function (trim, replace), to make understanding the code easier.

phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -1855,7 +1855,7 @@ public function format_property_text($value, $tag) {
...
```
alfsb, line 1855: The replace list here differs from other point. Maybe there is necessary two normalize functions, or if the difference is accidental, to test what list of substitutions is the correct.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -666,7 +666,7 @@ public function format_type_methodparam_text($type, $tagname) {
     }
 
     public function format_type_text($type, $tagname) {
-        $t = strtr(strtolower($type), ["_" => "-", "\\" => "-"]);
```
alfsb, line 669: Note how these substations lists are copied everywhere.

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -880,7 +880,7 @@ public function format_classsynopsis_oo_name_text($value, $tag) {
     }
 
     public function format_classname_text($value, $tag) {
-        if (($filename = $this->getClassnameLink(strtolower($value))) !== null && $this->cchunk["class_name_ref"] !== strtolower($value)) {
+        if (($filename = $this->getClassnameLink($this->normalizeFQN($value))) !== null && $this->cchunk["class_name_ref"] !== strtolower($value)) {
```
alfsb, line 883: This is somewhat unrelated, but in a future version, first test if `getClassnameLink()` with fails with raw string, and then test again with plain `\trim()`. In this case, generate the link as usual but generate an warning, as this a source error (unexpected whitespace around type/constante/etc tags).

PR #186, opened 2025-01-05
Define undefined properties and variables

haszi:
Define undefined properties and variables.

I'm planning on opening one or more PRs that would address some of the smaller issues found with static analysis before I try to make larger changes (such as adding parameter and return types). If anyone has any objections to this, please let me know.

### DIFF
--- a/phpdotnet/phd/Format.php
+++ b/phpdotnet/phd/Format.php
@@ -479,6 +479,8 @@ abstract class Format extends ObjectStorage
         return -1;
     }
     public function colspan(array $attrs) {
+        $to = 0;
+        $from = 0;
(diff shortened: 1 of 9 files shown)


Girgias (approved):
MSTM, but I wonder if we shouldn't be removing some of the packages/styles (I have no idea why PEAR would be used anymore)

haszi:
> MSTM, but I wonder if we shouldn't be removing some of the packages/styles (I have no idea why PEAR would be used anymore)

This would make sense. I'm not sure whether the libraries the PDF formats (Haru) and the format highlighters (GeSHi) are based on are still being maintained but those could probably be dropped.

PR #188, opened 2025-01-29
Add support enum links

SakiTakamachi:
In my local:

スクリーンショット 2025-01-31 17 27 57

Actually, it would be nice to be able to link `RoundingMode::HalfAwayFromZero` as well, but that might be a good idea to wait for a follow-up.

### DIFF
--- a/phpdotnet/phd/Package/PHP/XHTML.php
+++ b/phpdotnet/phd/Package/PHP/XHTML.php
@@ -118,6 +118,13 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
                 'classsynopsis' => 'format_classsynopsis_oo_name_text',
             ]
         ],
+        'enumname' => [
+            /* DEFAULT */ 'format_classname_text',
+            'oointerface' => [
+                /* DEFAULT */     'format_classname_text',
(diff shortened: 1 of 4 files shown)


SakiTakamachi:
@haszi 
With this change `enumname` behaves like `classname`, are there any concerns?
(I'm not very familiar with the docs)

phpdotnet/phd/Package/PHP/XHTML.php
```diff
@@ -118,6 +118,13 @@ abstract class Package_PHP_XHTML extends Package_Generic_XHTML {
                 'classsynopsis' => 'format_classsynopsis_oo_name_text',
             ]
         ],
+        'enumname' => [
+            /* DEFAULT */ 'format_classname_text',
+            'oointerface' => [
```
TimWolla, line 123: Should this be `ooenum`?

SakiTakamachi: yes, I forgot 😇

SakiTakamachi: fixed!

Girgias (approved):
Thank you!

PR #192, opened 2025-03-05
Render enumitemdescription using single line comments

sy-records:
Close https://github.com/php/doc-en/issues/4513

image
image


Girgias (approved):
This will work for the time being, thank you!

PR #195, opened 2025-03-06
Interactive annotations

derickr:
This goes hand in hand with https://github.com/php/web-php/pull/1228 and the changes I pushed to `doc-en` just now.


Girgias (approved):
MSTM, but I don't know which doc-en changes you are talking about?

PR #210, opened 2025-10-27
Make generated example IDs page specific

### LINKED ISSUE #205: Stabilize example IDs in HTML
Currently examples (specifically the `<div class="example"`) are assigned an id that looks something like `example-3066`. The number appears to be a global counter.

This results in the id assigned to a given example changing whenever any new examples are added before it anywhere in the manual.

These id's are useful tools for linking to specific examples via URL fragments (eg. https://www.php.net/manual/en/datetimeimmutable.add.php#example-2198 is, at the time of writing, Example #1 on the DateTimeImmutable::add documentation).

(While not immediately visible, they can be easily exposed via browser dev tools or extensions)

However, because they change relatively frequently, they're a lot less useful for any non-immediate use. For example, if someone's reading an older forum / Q&A answer that links to a specific example.

It would be useful if these id's could be stabilized. They only need to be unique on a specific page.

I suggest making the counter reset on a per-page basis - ie. the first example on any given page is always `example-1`, the next `example-2`.

AllenJB:
Fixes #205 

Makes the id's generated for examples reset for each page (except where they have another manually assigned id).

ie. the first example on any page is always `id="example-1"`, the second `id="example-2"` and so on (unless they have other id's manually assigned).

While hidden, these id's are often used for linking from articles, forums & Q&A sites around the web. Every time they change they break any links to a specific example.

An example of this can be seen on https://phpdelusions.net/pdo/fetch_modes#FETCH_BOUND - the bottom 2 sections of this article have links to specific examples, but these have long since broken due to the way id's are currently generated.

If you think there may be a better way to implement this (specifically where / how to implement the reset counter hook) please let me know.

### DIFF
--- a/phpdotnet/phd/Package/Generic/XHTML.php
+++ b/phpdotnet/phd/Package/Generic/XHTML.php
@@ -552,6 +552,8 @@ abstract class Package_Generic_XHTML extends Format_Abstract_XHTML {
 
     protected bool $exampleCounterIsPerPage = false;
 
+    protected array $perPageExampleIds = [];
+
     public function __construct(
         Config $config,
         OutputHandler $outputHandler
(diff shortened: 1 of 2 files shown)


phpdotnet/phd/Package/Generic/XHTML.php
```diff
@@ -2538,4 +2542,22 @@ public function format_whitespace($whitespace, $elementStack, $currentDepth) {
...
+            return $originalId;
+        }
+        if (preg_match('/^example\-[0-9]+$/', $originalId)) {
+            $this->perPageExampleCounter++;
+            return 'example-' . $this->perPageExampleCounter;
+        }
+        return $originalId;
+    }
+
+    public function onNewPage(): void
```
alfsb: `onNewPage()` should test `exampleCounterIsPerPage`, otherwise it will always reset, even with `exampleCounterIsPerPage` is false.

PR #212, opened 2025-11-03
Fix missing pages in search indexes

### LINKED ISSUE #211: Missing search (index) entries
I've noticed a significant number of entries don't appear in the web quick search. The entries for these are missing from the search indexes files.

Some examples are locale_canonicalize / Locale::canonicalize and XMLWriter::writePi / xmlwriter_write_pi.

In all the cases I found both the procedural and OO variants are missing, so it's not a case on one being skipped.

If you type 'XMLWriter' into the search, you can see the class entry comes up, as well as some methods. But the vast majority of methods are missing.

I've tracked this down to [the `$index['chunk']` check](https://github.com/php/phd/blob/3264920aa15a9cbc911aac80900e8b46356eeb50/phpdotnet/phd/Package/PHP/Web.php#L260). The offending entries all have `'chunk' = 0`.

However I don't know enough about what this `chunk` value is supposed to indicate, or what else might be using it, to take debugging this further.

One theory I have is that this is related to entries with multiple `<refname>` values in the same `<refentry>`.

AllenJB:
Fixes #211

I noticed there were significant pages missing from the web search.

After some investigation I found there were 2 circumstances where entries were missing.

In the first case, some entries were being excluded by the `!$index['chunk']` check. I found these could easily be re-included by skipping this check for specific `element` values. You can see the list of these running the following query on the `output/index.sqlite` file generated by a phd run:

```sql
SELECT *
FROM ids
WHERE chunk = 0
	AND element IN (
		'refentry',
		'stream_wrapper',
		'phpdoc:classref',
		'phpdoc:exceptionref',
		'phpdoc:varentry'
	)
```

Additionally entries were excluded where they all had the same docbook_id because of the way the `indexes` array is indexed by this id. Because the existing search indexes rely on this id, it's not easy to resolve. 

However, the current web search actually reworks these indexes to combine them anyway, and I'd created a pre-combined version of these indexes for #204 that does not rely on indexing by the docbook_id.

I've copied these changes to this PR and then modified them to pull the index list from the database without the deduplication. These will require further changes to php-web `js/search,js` and `js/search-index.php` - I've already made the changes to search.js for #204 (and the changes to `search-index.php` is just having it spit out the new -combined index file without any manipulation it currently does)

You can see the list of affected docbook_ids with the following query:

```sql
SELECT docbook_id
FROM ids
WHERE chunk = 1
	OR element IN (
		'refentry',
		'stream_wrapper',
		'phpdoc:classref',
		'phpdoc:exceptionref',
		'phpdoc:varentry'
	)
GROUP BY docbook_id
HAVING COUNT(*) > 1
```

Generally these are cases where there are both procedural and OOP interfaces. I'd guess which one makes it into the indexes is determined by the order in which the `<refname>` values appear.

There are some other cases such as stream wrappers (eg. bzip2:// and zlib://)

### DIFF
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -278,8 +278,8 @@ contributors($setup);
         $descriptions = [];
         foreach($this->indexes as $id => $index) {
             if (
-                (! in_array($index['element'], $alwaysIncludeElements, true))
-                && (! $index['chunk'])
+                (! $index['chunk'])
+                && (! in_array($index['element'], $alwaysIncludeElements, true))
             ) {
                 continue;
             }
@@ -316,8 +316,8 @@ contributors($setup);
         $indexes = $this->indexRepository->getIndexesWithDuplicates();
         foreach ($indexes as $index) {
             if (
-                (! in_array($index['element'], $alwaysIncludeElements, true))
-                && (! $index['chunk'])
+                (! $index['chunk'])
+                && (! in_array($index['element'], $alwaysIncludeElements, true))
             ) {
                 continue;
             }


phpdotnet/phd/Package/PHP/Web.php
```diff
@@ -254,10 +254,21 @@ protected function writeJsonIndex() {
...
+        ];
+
         $entries = [];
         $descriptions = [];
         foreach($this->indexes as $id => $index) {
-            if (!$index["chunk"]) {
+            if (
+                (! in_array($index['element'], $alwaysIncludeElements, true))
+                && (! $index['chunk'])
+            ) {
```
Girgias: I think having the `!index['chunk']` check first makes more sense as it should be quicker and also the most common reason for performing the second check?

PR #215, opened 2025-11-11
Add explicit type declarations to class constants (PHP 8.3+)

lacatoire:
### Summary
This PR introduces explicit type declarations for class constants, leveraging the feature introduced in PHP 8.3.

### Motivation
- Enhances type safety and static analysis.
- Makes code intent clearer.
- Prepares PhD for PHP 8.4 compatibility and future strictness.

### Scope
- Added types such as `int`, `string`, `bool`, or `array` to class constants.
- Example:
    Before:
        public const FORMAT_HTML = 1;
        public const NAME = 'PhD';
    After:
        public const int FORMAT_HTML = 1;
        public const string NAME = 'PhD';

### Impact
- ✅ No runtime behavior change.
- ✅ Fully backward compatible with PHP 8.3+.
- ⚙️ Code clarity and safety improvement only.

### References
- [PHP 8.3: Typed class constants RFC](https://wiki.php.net/rfc/typed_class_constants)
- [PHP manual: Class constants](https://www.php.net/manual/en/language.oop5.constants.php)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,11 +4,11 @@ jobs:
   ubuntu:
     strategy:
       matrix:
-          version: ['8.1', '8.2', '8.3']
+          version: ['8.1', '8.2', '8.3', '8.4', '8.5']
     runs-on: ubuntu-latest
     steps:
       - name: Checkout PhD
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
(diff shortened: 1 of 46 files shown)


Girgias (changes requested):
Can you update the composer.json file to require 8.3? I checked with Derick and the doc machine runs 8.3 so it should be fine.

Girgias (approved):

kocsismate:
@lacatoire At last, I had to revert these changes because it turned out that Gina got incorrect information, and the servers still run on PHP 8.2, and that's why the recent docs changes didn't synchronize to php.net.

cweiske:
The [Dockerfile in doc-en](https://github.com/php/doc-en/blob/master/.docker/Dockerfile) also still uses php 8.2.

PR #220, opened 2025-11-16
Add PHP 8.4 to test runner matrix

haszi:


haszi:
If there are no objections, I will merge this around the end of this week.

PR #223, opened 2025-11-25
Replace deprecated SplObjectStorage::attach() usage for PHP 8.5

lacatoire:
PHP 8.5 deprecates SplObjectStorage::attach().

Actual runtime message:
`"Method SplObjectStorage::attach() is deprecated since 8.5, use method SplObjectStorage::offsetSet() instead"`

PhD still calls these methods internally, producing E_DEPRECATED during builds.
This patch switches to offsetSet() on PHP >= 8.5 while keeping behavior unchanged on older versions.

### DIFF
--- a/phpdotnet/phd/ObjectStorage.php
+++ b/phpdotnet/phd/ObjectStorage.php
@@ -3,23 +3,18 @@ namespace phpdotnet\phd;
 
 class ObjectStorage extends \SplObjectStorage
 {
-	public function attach(object $object, mixed $info = array()): void {
+	public function attach(object $object, mixed $info = []): void {
 		if (!($object instanceof Format)) {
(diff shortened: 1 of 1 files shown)


phpdotnet/phd/ObjectStorage.php
```diff
@@ -16,6 +16,12 @@ public function attach(object $object, mixed $info = array()): void {
 			);
 		}
 		parent::attach($object, $info);
+
+        if (PHP_VERSION_ID < 80500) {
+            parent::attach($object, $info);
+        }
```
haszi: `SplObjectStorage::offsetSet()` seems to be available since PHP 5.3. I think it's safe to drop this fallback completely.

PR #228, opened 2025-12-15
Fix: Prevent using null as array offset in Package/PHP/Web

### LINKED ISSUE #227: Using null as an array offset is deprecated
https://github.com/php/phd/blob/5d705c2676cbc85d9f94e212e796f776fcbf59fc/phpdotnet/phd/Package/PHP/Web.php#L207
Trying to access array offset on null
https://github.com/php/phd/blob/5d705c2676cbc85d9f94e212e796f776fcbf59fc/phpdotnet/phd/Package/PHP/Web.php#L440

lacatoire:
Addresses the deprecation warning/error related to attempting to use a `null` value as an array.

This was primarily fixed in `phpdotnet/phd/Package/PHP/Web.php` by using `($this->property ?? [])` to ensure that the array access operator (`[]`) is always applied to a valid array, thus resolving the issues experienced with modern PHP versions (7.4+).

[Fixes #227](https://github.com/php/phd/issues/227)

### DIFF
--- a/phpdotnet/phd/Package/PHP/Web.php
+++ b/phpdotnet/phd/Package/PHP/Web.php
@@ -205,7 +205,13 @@ $PARENTS = ' . var_export($parents, true) . ';';
             "source" => $this->sourceInfo($id),
         );
         $history = $this->history ?? [];
-        $setup["history"] = $history[$setup["source"]["path"]] ?? [];
+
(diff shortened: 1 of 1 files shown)


plusminmax:
Please pardon my english, since it is not my default tongue.
If `sourceInfo` return `null`
then `setup source is null`
thus `setup source path` access on null.

PR #229, opened 2025-12-18
Switch autoloading from PSR-0 to PSR-4

lacatoire:
This PR replaces the deprecated PSR-0 autoload configuration with a minimal PSR-4 setup.

No code is moved, no files are added, and there are no functional or behavioral changes.
This follows the idea introduced in https://github.com/php/phd/pull/52 , but intentionally reduces the scope to a minimal, infrastructure-only change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,8 +27,13 @@
         "easybook/geshi": "For using GeSHi as code highlighter"
     },
     "autoload": {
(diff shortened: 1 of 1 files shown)


Girgias (approved):
Thanks for taking care of this!

PR #249, opened 2026-03-11
Fix rendering of attribute constants

kocsismate:
$this->cchunk["fieldsynopsis"]["modifier"] was not set for atrributes due to the early return, but this value is needed in order to be able to decide if the fieldsynopsis refers to a constant or a property at https://github.com/php/phd/blob/d7f700463918a466705e92b4bfd4df0afa39d944/phpdotnet/phd/Package/Generic/XHTML.php#L1608


Girgias:
Could you add a test file? :)

DanielEScherzer (commented):
makes sense as far as I understand it, but I don't fully understand it and haven't interacted with the phd system before, so not confident enough to actually approve it

Girgias (approved):

PR #252, opened 2026-05-11
GH-248: Add `function` keyword to `<methodsynopsis>`.

### LINKED ISSUE #248: `<methodsynopsis>` element should render the `function` keyword
Just realised that we don't mention the `function` keyword when rendering.

Many people have previously complained that our synopsises don't look like valid PHP code and we've been slowly remediating this. I guess this is another instance of a rendering change we should implement.

jordikroon:
Fixes #248

Not sure if this needs additional tests, since the existing tests pretty much cover it all. WDYT?

---
Function:

image

---
Class:

image


kocsismate (approved):

Girgias (approved):
I think the existing tests are sufficient

PR #253, opened 2026-06-07
Remove unused historyFile config property

jordikroon:
This property doesn't seem to be used anymore.


alfsb:
This is related to https://github.com/php/doc-base/blob/master/configure.php#L1079 ?

alfsb:
I asked because I'm interessed in moving all PhD stuffi into a separate `genphdfiles.php`, so doc-base/configure.php can be faster, by focusing only in joining and validating the XML files. In fact, I'm interessed in moving all PhD code into PhD repo.

